PR #186, opened 2015-01-27
allow reindexing after inserting items in the database

naeemmuhammad:
We are facing a problem with indexing while adding or removing set items (objects) from a set, this fix solves a part of that problem. 
When we add or delete a set item from a set, item is removed from the database but its index information is not updated. Adding to set from object editor is a problem, whereas it works fine when an object is added to a set from set editor. Problem with adding items to a set occures only when added from object editor, whereas it works fine when added from set editor. 
This particular fix solves the problem related to 'adding an item'.  Currently, in add (insert) operation new set item is added to the database and indexed properly, however index of the object is not updated, this is being caused by the false value of the reindex mode. Setting reindex mode to true solves the problem.

We are working on 'delete item' part of the problem and will make another pull request as soon as we have a solution.

More information about the problem can be found on our question on Collective Access forum:
http://www.collectiveaccess.org/support/forum/index.php?p=/discussion/294068/set-item-removal-problem

### DIFF
--- a/app/lib/core/BaseModelWithAttributes.php
+++ b/app/lib/core/BaseModelWithAttributes.php
@@ -524,7 +524,7 @@
 				// set the field values array for this instance
 				$this->setFieldValuesArray($va_field_values_with_updated_attributes);
 				
-				$this->doSearchIndexing($va_fields_changed_array);
+				$this->doSearchIndexing($va_fields_changed_array, true);
 				
 				if ($vb_web_set_change_log_unit_id) { BaseModel::unsetChangeLogUnitID(); }
 				if ($this->numErrors() > 0) {


collectiveaccess:
You can't enable reindexing like that across the board unfortunately. It makes indexing performance much worse, and it's already too slow.

There has been a fix for the set indexing issues in the develop branch for a while that addresses the underlying issue in the incremental indexer. You might want to try that.

naeemmuhammad:
I tested the development version, this problem exist there as well. I am looking for another solution without performance overheads, as soon as i have one i will get your feedback on that.

collectiveaccess:
I've testing adding and remove objects from sets from both the set editor and the "sets" bundle in the object editor with the current "develop" code, and it works fine. I'm looking at the index directly as I'm doing this, and the entries are correctly being created and removed. 

If you're testing the indexing by performing a search, and you have search caching enabled, you may be seeing unexpected results because of cache lag.

naeemmuhammad:
Perhaps there is a difference between our test environments, mainly between the search engines. We tested it with solr, which does not support incremental indexing. And indexing in providence is performed differently for systems with and without incremental indexing support. Maybe this problem occurs only for search engines without incremental indexing support.

Disabling search caching did not make any difference. 

I was wondering if there could be a problem in our configurations, perhaps it is worth asking you to have a look at our search_indexing.conf file:
https://github.com/libis/providencesandbox/blob/master/app/conf/search_indexing.conf

PR #187, opened 2015-01-28
update indexing for related rows after deletion

naeemmuhammad:
This pull request is in the context of our previous pull request (https://github.com/collectiveaccess/providence/pull/186), where we described the indexing issue we are facing with sets. After removing an item from a set index is not updated for all related tables. 
For example, if we remove an object from a set, index needs to be updated for ca_sets and ca_objects, however index for ca_objects is not updated. This fix solves this problem by triggering indexing for the remaining related table.


naeemmuhammad:
I was wondering if there is already some feedback on this fix. Thanks

naeemmuhammad:
Any feedback on this pull request will help us to understand if this fix can be used for the above mentioned problem.

collectiveaccess:
Try using the latest master-fix code. I believe the issue was solved there a while back.

naeemmuhammad:
Some updates on this pull request:
We found that master-fix too has the same issue. After a lot of in-house testing we are now using this fix in our collective access 1.4 systems successfully.

PR #191, opened 2015-02-27
Added some top margin to the first related item

jasper-d:
Before/after:

margin-top_before_after


collectiveaccess:
Thank you!

jasper-d:
That was fast :)

Welcome & thanks for pulling it.

PR #192, opened 2015-03-02
Remove duplicate index that causes a setup error

jasper-d:
An index is defined two times, causing an error during setup.


collectiveaccess:
This wasn't causing errors on the systems we are using. But it's clearly wrong. Thanks for fixing!

PR #193, opened 2015-03-04
PROV-1185 Redis LRU cache support

jasper-d:
It's actually a trivial addition, but since the old cache abstraction of Doctrine isn't compatible with some RedisPHP builds (cf. https://github.com/doctrine/cache/commit/49acabe477b35b306a10d730ddb3fa9586c60982#diff-814870644de02e5bca3df56166aaca53) I had to update Doctrine/common. Turned out that Doctrine/cache was moved in its own module, therefore I removed doctrine/common and added doctrine/cache. I couldn't find any other code that depends on doctrine/common (hope I didn't miss anything).

I checked that file cache, memcached and Redis (still) work under Linux and Redis and file cache under Windows too.

A possible improvement would be support for Unix socket connections, but don't think that this is really needed.

PS: 12 unit tests are failing before and after the update.


skeidel:
There shouldn't be any other doctrine dependencies so updating is probably fine. I'll give it a try later today. Thanks!

Oh and yes, some of the tests are broken in develop. We're working on that - in a different branch. The important thing is that the cache unit tests pass with the new backend.

skeidel:
Ok the Redis part worked very well and memcached and apc survived the doctrine update too. It broke the filesystem based cache though, but this had nothing to do with your code. They just changed their API a little bit. 

I fixed it and merged your code. Thanks!

jasper-d:
Thanks, didn't notice that file system cache wasn't working, sorry about that.

PR #196, opened 2015-03-19
Grey background for disabled input elements

jasper-d:
Just a minor usability improvement to help users recognize read-only fields. #F0F0F0 is default for FF and IE, slightly different from Chromes default bg color for disabled input elements (#EBEBE4).

disabled_background


skeidel:
I could have sworn I did that a while back. Well, thanks!

PR #197, opened 2015-03-19
Fix search/browse criteria CSS

jasper-d:
Just a little fix for search/browse criteria that span multiple lines.
criteria_css

### DIFF
--- a/themes/default/css/base.css
+++ b/themes/default/css/base.css
@@ -3936,6 +3936,7 @@ div.quickSearchNoResults {
     text-transform:uppercase;
     font-size:11px;
     line-height: 27px;
+    white-space: nowrap;
 }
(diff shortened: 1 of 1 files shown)


skeidel:
Looks good, except that the "Start Over" button breaks in two pieces when the existing criteria align just right:

screen shot 2015-03-19 at 08 51 26

Now that might have nothing to do with your fix but I just noticed it :-)

jasper-d:
I think that has nothing do to with my changes, fixed it anyway. :-)

skeidel:
Looks good, thanks! We like it when other people fix our problems.

jasper-d:
Welcome, I like if I don't have to write software from scratch.

PR #198, opened 2015-04-15
caUtils requires initializeLocale.php

docudoc:
Couldn't run caUtils without requiring the initializeLocale class.

### DIFF
--- a/bin/support/caUtils
+++ b/bin/support/caUtils
@@ -1,0 +1,184 @@
+#!/usr/bin/env php
+<?php
+	if (!_caUtilsLoadSetupPHP()) {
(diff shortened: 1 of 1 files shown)


skeidel:
Why did you add a new file? The caUtils script is in support/bin

Also, what exactly happened when you tried to run it? caUtils includes setup.php, which includes app/helpers/preload.php which in turn includes initializeLocale:

https://github.com/collectiveaccess/providence/blob/master/app/helpers/preload.php#L41

docudoc:
Uhm. Sorry about that. I did think it a bit strange that I ended up with a new file.

I got the an CLI error message saying that initializeLocale wasn't defined. I added the require, and it ran fine.

skeidel:
Did you by any chance remove the require_once statement for preload.php in setup.php?

PR #201, opened 2015-04-30
PROV-1240 Travelogue plugin has to return truthy

kehh:
* If it doesn't return true then subsequent plugins implementing hookPeriodicTask get disabled.


skeidel:
That branch name is quite interesting ...

kehh:
hey git didn't limit me ;)

PR #206, opened 2015-06-03
PROV-1268 Add token-based authentication to web service API

skeidel:
This finally allows us to re-use sessions for the API. While this is technically not REST it's common practice and allows us to not basic auth for every request.

The idea is that you call the /service.php/auth/login endpoint via GET (w/ basic auth) and it'll give you an authToken. Append this authToken as GET parameter to any subsequent service API calls and you'll be authenticated and find yourself in the same session.

Auth tokens don't last very long (which is kind of the point) so you'e going to have to re-authenticate every once in a while, when you get a 401.

The old "basic auth on every request" way still works for compatibility.

A good next step would be to allow users to generate API keys that can be used for service authentication instead of their account passwords.

I updated the PHP API wrapper at https://github.com/skeidel/ca-service-wrapper to support this and will update the readme soon.

### DIFF
--- a/app/conf/access_restrictions.conf
+++ b/app/conf/access_restrictions.conf
@@ -656,7 +656,7 @@ access_restrictions = {
 			actions = { can_delete_ca_storage_locations }
 		}
 	},
-	
+
 	editor/loans/LoanEditorController = {
 		default = {
 			operator = OR,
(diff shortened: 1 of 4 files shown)


gautiermichelin:
Nice, that would definitely be of some use.

PR #207, opened 2015-06-14
NOJIRA : updating locale files with new strings

gautiermichelin:
- app/locale/en_US/messages.po : merging new strings inside i18n file, messages.po is used inside CA not as a regular po file, but more as a pot catalog file
- locales po/mo files : integrating new strings through POEdit "update strings from a POT file"
- app/refineries/objectRepresentationSplitter/objectRepresentationSplitterRefinery.php & app/models/ca_editor_ui_screens.php : _('') is a bad idea when we need to update the gettext catalogs...


gautiermichelin:
Thanks Stefan.

PR #208, opened 2015-06-15
NOJIRA : improving french translation

gautiermichelin:


skeidel:
Did you change anything? I pulled the latest version from Transifex yesterday.

gautiermichelin:
Yes, I do my translation on PoEdit, through regular pull requests... and 
transifex was not at the point of master in github at all :-(

Now it's up to date, and my PR is in sync with transifex, but it wasn't.

skeidel:
Ok. Transifex should now pull from master-fix automatically. So whenever you want to update the Transifex source, just create a PR with an updated en_US messages.po.

PR #214, opened 2015-07-30
NOJIRA configurable disabled items in hierarchy browser

kehh:
PROV-1193 inadvertently removed all the disabled list items in hierarchy browsers. This PR keeps the (new) default behaviour for hiding the list items and makes this configurable in app.conf through some keys. These can be set to `disabled`, `hide` and `full`.


skeidel:
Making this configurable makes sense and doesn't hurt but I'm wondering why you'd want to show disabled items in `hierarchy_location` and `hierarchy_navigation`? :confused:

kehh:
We have a  taxonomic hierarchy as ca_lists and there are disabled parent terms. If they are hidden the hierarchy browser looks weird with empty columns. `disabled` actually works quite well for this as you can then restrict the user to only lower level terms. Also we show the hierarchy_navigation on item editor pages.

leftclickben:
An example use case is a storage location  hierarchy with rooms, shelves and boxes. By displaying but disabling the rooms and shelves, the user is forced to specify the location to the box level.

PR #215, opened 2015-07-30
LDAP authentication refactor

leftclickben:
This is a refactor of the new authentication code. There are three outcomes:

1. It actually works for ActiveDirectory configurations.
2. The post-authentication HTTP behaviour is fixed (see 0ce6f51 for details).
3. Improvements to the code style.

The major code style improvement is changing `IAuthAdapter` from a "static interface" to a normal interface that defines non-static methods. This means that the implementations have also changed to normal classes with instances, and `AuthenticationManager` holds a reference to an instance of the applicable implementation and calls methods on that instance, rather than calling static methods on a class identified by name.

I also created `AbstractLDAPAuthAdapter` which contains the LDAP functionality that is common between OpenLDAP and AD (which is most of it).

### DIFF
--- a/app/lib/core/Auth/AbstractLDAPAuthAdapter.php
+++ b/app/lib/core/Auth/AbstractLDAPAuthAdapter.php
@@ -1,0 +1,306 @@
+<?php
+/** ---------------------------------------------------------------------
+ * app/lib/core/Auth/AbstractLDAPAuthAdapter.php : Abstract base class for LDAP adapters
+ * ----------------------------------------------------------------------
+ * CollectiveAccess
+ * Open-source collections management software
+ * ----------------------------------------------------------------------
+ *
+ * Software by Whirl-i-Gig (http://www.whirl-i-gig.com)
+ * Copyright 2014 Whirl-i-Gig
+ *
+ * For more information visit http://www.CollectiveAccess.org
+ *
+ * This program is free software; you may redistribute it and/or modify it under
+ * the terms of the provided license as published by Whirl-i-Gig
+ *
+ * CollectiveAccess is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTIES whatsoever, including any implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * This source code is free and modifiable under the terms of
+ * GNU General Public License. (http://www.gnu.org/copyleft/gpl.html). See
+ * the "license.txt" file for details, or visit the CollectiveAccess web site at
+ * http://www.CollectiveAccess.org
+ *
+ * @package CollectiveAccess
+ * @subpackage Auth
+ * @license http://www.gnu.org/copyleft/gpl.html GNU Public License version 3
+ *
+ * ----------------------------------------------------------------------
+ */
+
+require_once(__CA_LIB_DIR__.'/core/Auth/BaseAuthAdapter.php');
+require_once(__CA_LIB_DIR__.'/core/Auth/PasswordHash.php');
+
+abstract class AbstractLDAPAuthAdapter extends BaseAuthAdapter {
+    private $opo_auth_config;
+    private $opo_ldap;
+    # --------------------------------------------------------------------------------
+    public function __construct() {
+        if (!function_exists("ldap_connect")){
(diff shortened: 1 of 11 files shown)


leftclickben:
PLEASE NOTE:  We are not able to test either the OpenLDAP implementation or the ExternalDB implementation.  I don't see any cause for concern about the ExternalDB one but please check that OpenLDAP works -- it should but without actually trying it, we can't be certain.  ActiveDirectory and CaUsers definitely work.

app/lib/core/Auth/AbstractLDAPAuthAdapter.php
```diff
@@ -0,0 +1,306 @@
...
+            'active' => $this->getConfigValue("ldap_users_auto_active"),
+            'roles' => array_merge($this->getConfigValue("ldap_users_default_roles", array()), $this->getRolesToAddFromDirectory($ps_username)),
+            'groups' => array_merge($this->getConfigValue("ldap_users_default_groups", array()), $this->getGroupsToAddFromDirectory($ps_username))
+        );
+    }
+    # --------------------------------------------------------------------------------
+    public function createUserAndGetPassword($ps_username, $ps_password) {
+        // We don't create users in directories, we assume they're already there
+
+        // TODO FIXME The following is insecure!
```
leftclickben: It is insecure in the following scenario:

1. User exists in LDAP and is given appropriate security group assignments to allow them access to CA.
2. User logs in to CA; the code here hashes the password and saves it in the CA database; also the roles are synced.
3. User is removed from the LDAP security group.  This has previously been all that is required to prevent that user from accessing CA.
4. Now that the `AuthenticationManager` is trying the configured adapter, and then falling back to the `CaUsers` implementation, *after the user has been removed from CA via LDAP they can still login*.  Their password still works because we saved the hash that was created here.

Actually the main source of bad security here is the fallback to `CaUsers` in `AuthenticationManager`.  Previous implementation had an explicit list of users that bypass LDAP; that was secure.

When a user is removed from the LDAP group that gives them access to CA, they should no longer be able to login to CA.

leftclickben:
Ideally we would like this code to be incorporated into a 1.5.1 release.  I think this is reasonable given that AD simply does not work with 1.5.

app/lib/core/Auth/AbstractLDAPAuthAdapter.php
```diff
@@ -0,0 +1,306 @@
...
+    }
+    # --------------------------------------------------------------------------------
+    public function supports($pn_feature) {
+        switch($pn_feature){
+            case __CA_AUTH_ADAPTER_FEATURE_AUTOCREATE_USERS__:
+                return true;
+            case __CA_AUTH_ADAPTER_FEATURE_RESET_PASSWORDS__:
+            case __CA_AUTH_ADAPTER_FEATURE_UPDATE_PASSWORDS__:
+            default:
+                return false;
```
leftclickben: Note this has changed for AD, previously `__CA_AUTH_ADAPTER_FEATURE_RESET_PASSWORDS__` and `__CA_AUTH_ADAPTER_FEATURE_UPDATE_PASSWORDS__` were `true`, but it doesn't make sense to allow a user to change password when they are authenticating with an external Directory.  (OpenLDAP implementation was already returning `false` for both of these).

skeidel:
We have a bunch of users on OpenLDAP and almost none running AD, so we should definitely make __sure__ this works.

And yes, agreed about the CaUsers fallback. I had a user complain about that just recently.

skeidel:
Oh yeah, and this may have to wait until Seth gets a chance to look at it because he's dealing with most of the client systems that actually use this. That's probably not going to happen until next week.

leftclickben:
Thanks @skeidel, no worries on timing, we have merged this to our fork already as we needed it working on AD to move to 1.5. 

I've tried to keep the OpenLDAP interactions unchanged, just refactored the code structure, but if there are any issues, feel free to push up fixes to this branch if it makes things easier.

skeidel:
A client found an issue where logging without a password was possible when using OpenLDAP. Apparently ldap_bind() will bind successfully if no password is provided! I'd imagine the same is true for ActiveDirectory.

I used that opportunity to test this PR with OpenLDAP and it looks ok.

PR #216, opened 2015-07-30
Fix circular dependency.

leftclickben:
+ `BaseObjectLocationModel` was `require`ing `ca_movements` and `ca_movements` was `require`ing `BaseObjectLocationModel` -- circular dependency.
+ It was not necessary to `require` `ca_movements` from `BaseObjectLocationModel` so this has now been removed.


kehh:
This causes a fatal error when using xcache  hence the need for this change.

PR #217, opened 2015-07-30
NOJIRA : Disabling cli colors under windows

gautiermichelin:
Color are not handled under Windows : http://d.pr/i/1l2sU
capture d ecran 2015-07-30 a 20 55 51


skeidel:
You may want to use the [OS family helper](https://github.com/collectiveaccess/providence/blob/master-fix/app/helpers/utilityHelpers.php#L472) we already have instead?

skeidel:
Or [this one](https://github.com/collectiveaccess/providence/blob/master-fix/app/helpers/utilityHelpers.php#L488)

gautiermichelin:
Sure.

gautiermichelin:
Done with the first one.

PR #223, opened 2015-08-31
NOJIRA : allows to hook TimeExpressionParser preprocess

gautiermichelin:
This allows me to develop into a plugin a preparser for French Revolutionary Calendar dates.
The plugin would allow me to parse dates like "12 thermidor an II" (https://en.wikipedia.org/wiki/French_Republican_Calendar) before sending a normal date to the TEP class. This is only for input now.

In the future, it would allow the development of plugins for date entering from Hebrew, Islamic, persian, chinese calendars, probably with a special text mark to distinguish the calendar if the month names don't help. Going through plugins seems me a prudent way of digging into those.

### DIFF
--- a/app/lib/core/Parsers/TimeExpressionParser.php
+++ b/app/lib/core/Parsers/TimeExpressionParser.php
@@ -687,6 +687,14 @@ class TimeExpressionParser {
 	}
 	# -------------------------------------------------------------------
 	private function preprocess($ps_expression) {
+
+		// Trigger TimeExpressionParser preprocess hook
+		$o_app_plugin_manager = new ApplicationPluginManager();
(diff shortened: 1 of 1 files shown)


skeidel:
Are you sure you want to do this *after* the preprocess function has mangled the input to pieces? I'd do it before (or add hooks for both, I guess?).

gautiermichelin:
Hi Stefan,

I'm better with doing this after, so ordinals like "1er" are treated ; 
accents, capitals are removed ; wordsWithSpaces are handled, and so on.
But I'm ok with doing both hooks naturally.
I'm adding this fast and commiting back.

Gautier

gautiermichelin:
done with 2 hooks, 1 before, 1 update, named in consequence.

gautiermichelin:
Thanks.

PR #225, opened 2015-09-30
NOJIRA - Make it possible to have a theme other than the default theme

kehh:
- Without this the it looks for an override with a path like objects/editor/screen_html.php not editor/objects/screen_html.php and also grumbles that it cannot find the view path.


kehh:
Here's the error message you get without this fix.
nooverridetheme

kehh:
Closing this as it's implemented in PROV-1794 (thanks for spotting that @leftclickben)

PR #227, opened 2015-10-08
NOJIRA : allow xml display when clicking on a XML representation

gautiermichelin:
If you click on a XML the media, the caMediaDisplayContent can't as htaccess blocks it.


gautiermichelin:
Thanks

PR #237, opened 2016-01-07
PROV-1443 Fix Too many permissions make roles not save

kehh:
* Use LONGTEXT for `vars` and `field_access` columns in ca_user_roles

This fixes http://clangers.collectiveaccess.org/jira/browse/PROV-1443

### DIFF
--- a/app/version.php
+++ b/app/version.php
@@ -3,7 +3,7 @@
 	define('__CollectiveAccess__', '1.5.2');
 	
 	# Schema revision
-	define('__CollectiveAccess_Schema_Rev__', 121);
+	define('__CollectiveAccess_Schema_Rev__', 122);
 	
 	# Release type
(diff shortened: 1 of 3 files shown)


gautiermichelin:
+1 : nice kehh, seen the problem here with multiple plugins

kehh:
I was unsure whether to use `mediumtext`, but the rest of the schema uses `longtext`? As @leftclickben says (https://github.com/wamuseum/providence/pull/218) it does seem a bit of an overkill.

collectiveaccess:
The difference in longtext vs. mediumtext is an extra byte for a field in a table that might have ~200 rows at the outside. It really doesn't matter which is used, but I'd err on the side of consistency and use longtext.

leftclickben:
We decided the same thing.

kehh:
200 roles! I thought we were being a bit overzealous with 45.

>

PR #241, opened 2016-02-10
Fix display template issue

leftclickben:
+ When processing related records, link (`<l>`) tags
  were not working correctly, then never generated an
  HTML link (`<a>`) tag.
+ This fixes the issue by passing the related table and
  primary key, instead of the relationship table and
  primary key.

IDE also auto-fixed a bunch of whitespace, append `?w=1` to the github URL to ignore it.

### DIFF
--- a/app/helpers/displayHelpers.php
+++ b/app/helpers/displayHelpers.php
@@ -2742,7 +2742,7 @@ define("__CA_BUNDLE_DISPLAY_TEMPLATE_TAG_REGEX__", "/\^([0-9]+(?=[.,;])|[\/A-Za-
 				} else {
 					switch(__CA_APP_TYPE__) {
 						case 'PROVIDENCE':
-							$va_links[] = ($vs_link = caEditorLink($g_request, $vs_text, $ps_class, $ps_table_name, $pa_row_ids[$vn_i])) ? $vs_link : $vs_text;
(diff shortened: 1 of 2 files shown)


collectiveaccess:
Thanks for submitting this. A problem I can see is that by shifting the context of the template to the related item you're precluding display of "interstitial" data (data bound to the relationship). More than a few users have models that rely on interstitial data. 

You can get <l> to work without this change by using a <unit relativeTo> in your template that shifts context to the related tables. I realize this is clunky, and thought should be given to implementing magic behavior for <l> allowing it to shift context for linking purposes only.

leftclickben:
Yeah, I thought this change would be a bit too far-reaching.  The methods I changed are both used fairly widely so I didn't know what other effects it might cause.

One solution is to add an optional attribute to the `<l>` element, e.g. `<l to="interstitial">` (the default, equivalent to `<l>`) or `<l to="related">`.  Just a thought, but I don't plan on implementing that.

Also, sorry, but I think something went wrong with github processing your comment, because something is missing:

> You can get to work without this change by using a in your template that shifts context to the related tables.

What can I use in the template as a workaround?

leftclickben:
I figure you mean to use `<unit relativeTo="ca_entities">`.  I think part of this pull request is still worthwhile, I'll put up another PR.

PR #244, opened 2016-03-16
NOJIRA fix fatal error class ca_metadata_elements is not defined

kehh:
Fixes for PROV-1566 seem to have broken quick add, and also possibly the hierarchy browser due to undefined class `ca_metadata_elements`.

I've added the include in `__construct()` as a number of other files are included there.


kehh:
Looks like the various Getty LOD driven tests failed. The services seem to be pretty unreliable I don't have a 'build now' button in my travis for collectiveaccess/providence. I don't think the failure relates to this current change.

stefankeidel:
Yup, Getty has been acting up all week. Thanks for the fix!

PR #245, opened 2016-03-22
Support HTTPS

leftclickben:
Previously the protocol was assumed to be plain HTTP, and a change to `global.conf` was required to support HTTPS.

This change detects the protocol used for the request and uses that as the base application protocol.  This is similar to the way the hostname is detected.

The "other way" is to redirect HTTP->HTTPS but I don't see why port 80 should even be open for this internal system.  We can have that discussion but either way, it shouldn't be constantly bouncing back to HTTP and then back to HTTPS, especially when the requests that seem to do this are around the login process, where passwords are involved, i.e. exactly when you don't want to be dropping out of HTTPS.

On the staging and uat servers, I have hacked the current deployment to change `global.conf` to hardcoded "https" instead of "http".


leftclickben:
Whoops!  Opened the pull request to the wrong branch, sorry.  I will submit this change upstream but this PR was supposed to be to our fork.

stefankeidel:
Sounds like a useful change though! I find myself changing the protocol in global.conf all the time, which is not good. http -> https redirects should be left to the webserver config imo.

leftclickben:
I'll be pushing a couple of things your way in the next day or two, including this (but on the right base branch!) and something a bit bigger.  

And yes agree the redirect should be at the web server level, my comment on that was more for @kehh about our particular servers.

PR #255, opened 2016-06-03
Bug solving in source id restrictions for browsing and accessing object through URL

albertocpbranco:
Bug solving in source id restrictions for browsing and accessing object through URL


stefankeidel:
Can you base this off of the `master-fix` branch, please? `master` is always the latest release code

PR #256, opened 2016-06-03
Bug solving in source id restrictions for browsing and accessing object through URL

albertocpbranco:
Bug solving in source id restrictions for browsing and accessing object through URL


stefankeidel:
Thanks! Will take a closer look at this on Monday

PR #258, opened 2016-06-10
PROV-1727 Add global view for change log

stefankeidel:
There are two different permissions: `can_view_own_change_logs` lets a user view their own changes under Manage > Logs > Change Log, while `can_view_change_logs` allows for a global view over all changes.


stefankeidel:
Oops, wrong branch

PR #262, opened 2016-06-21
NOJIRA Add trim and join support to expression parser

kehh:
We needed this and have added it. If this gets merged I'll document in the Wiki.


kehh:
Thanks

PR #263, opened 2016-07-18
NOJIRA Add `daterangepicker` in `browsable` assets class

kehh:
* Fixes an undefined jQuery().daterangepicker when opening an advanced search form with a date element in it

Feel free to decline this PR if this is better fixed elsewhere.


collectiveaccess:
I'm going to do this in a slightly different way...

PR #267, opened 2016-08-10
Test query builder

collectiveaccess:


kehh:
@CollectiveAccess Just a quick heads up that this contains a bunch of unrelated changes including a customised navigation plugin, some theme tweaks, some custom print templates and some other changes on top of the query builder,  so you probably don't want it to land in a release.

collectiveaccess:
Yeah, I just merged it into a develop-based branch to see how much broke. The merge was pretty easy and seems fine so far...

PR #268, opened 2016-08-17
Fix DOCX export in PHP7

kehh:
* `String` is now a required word in PHP 7, and phpword now has this fixed in https://github.com/PHPOffice/PHPWord/releases/tag/v0.13.0
 - As a result of this DOCX export fails spectacularly in php7 with a 500
* phpword now requires being installed via composer (good thing TM)
* Remove phpword from `__CA_LIB_DIR__."/core/Parsers/PHPWord`
* use htmlentities for all content being inserted into the document (`&` was breaking output)

(did I mention, it would be really good to not include vendor code in git)


app/lib/ca/BaseFindController.php
```diff
@@ -742,8 +742,6 @@ protected function _genExport($po_result, $ps_output_type, $ps_output_filename,
 						$this->render('Results/xlsx_results.php');
 						return;
                     case '_docx':
-                        require_once(__CA_LIB_DIR__."/core/Parsers/PHPWord/Autoloader.php");
```
kehh: phpword now uses the composer autoloader

kehh:
Some sample files obtained using php 5.6:
(there shouldn't be any significant difference between these files).
Before:
[Export (27).docx](https://github.com/collectiveaccess/providence/files/421965/Export.27.docx)
After:
[Export (26).docx](https://github.com/collectiveaccess/providence/files/421964/Export.26.docx)

kehh:
I just noticed that this has been worked around in master-fix, but phpoffice now has a stable release with the PHP7 fix in it. I'd propose that this PR gets included in favour of the current changes in master-fix (https://github.com/collectiveaccess/providence/commit/66521b0d6342859e4b8f13c1f83465faeebf8cc5 and https://github.com/collectiveaccess/providence/commit/02152f8f02df10cc1051ebd71bfc3f4bac407ef5)

collectiveaccess:
Thanks for taking care of this! 

I agree about using Composer for vendor code. The suck-it-in-and-commit practice started long before Composer and the like existed and is clearly obsolete. In the not-too-distant future I'd like to remove other dependencies to Composer, or remove them entirely. Similarly, the javascript assets system is a mess and should be moved out into NPM/Bower/et al land. We should discuss this whenever we manage to have that meeting.

gautiermichelin:
Thanks for that, moving to Composer would be a huge step forward.
Before you decide for JS, I'd say +1 for NPM instead of Bower.

leftclickben:
@gautiermichelin They both have their strengths and weaknesses, any particular reason you chose npm over bower?

collectiveaccess:
I'm kind of leaning towards NPM, if only because I'm used to it from another (non-CA related) project and because I tend to agree with the points made in articles like https://gofore.com/stop-using-bower/

I'm very open to being told I'm wrong tho' :-)

gautiermichelin:
Hi Ben,

I've used npm on quite a daily/weekly base for an angular JS project for a zoo gaming zone. I also have been using node.js Ghost blogging platform for a while, both using npm.

I've used a bit bower in the past.

So absolutely no tech consideration, just a good feeling of the last experiences.

:-)

leftclickben:
I've been pretty busy for the last month so I haven't had a chance to get back to this.  I'm primarily a javascript developer so this is something I've thought about over time for my own stuff as well.

Most of the points made in the linked article are either spurious or outright wrong.  The first section, "what is wrong with bower", is actually what is _right_ with bower...

In the context of node.js, it makes sense to allow each package to depend on its own version of a particular other package.  There is no cost to having multiple different versions of the same library, because there is no runtime code download going on.  In fact, in earlier versions of npm, it installed a different copy for every time it is listed as a dependency, even if the same version was already installed somewhere else (nowadays, it attempts to make `node_modules` as flat as possible, and only nests when there is a version difference).

However, in a browser, you don't want to include the same library twice if you can avoid it.  So that warning box that is in the image, is exactly what you want to happen.  You resolve the conflict in a development setting, and you specify the version you want in your `bower.json`.  If there is an incompatibility, you want to know about that sooner.

On some of the other points -- "when github gets DDOSed, your build breaks" is spurious.  What about when npm gets DDOSed?

The point about module loading is also not exactly correct.  I've never used `wiredep` but it looks quite hideous.  A much better solution involves `require.js` with `bower` and `gulp`.  Or if you prefer, you can still use `browserify` with `bower`.  Most decent modules use a UMD pattern these days, exposing both a CommonJS module and an AMD module.

One thing that isn't mentioned, which is relevant to a case like this, is availability of packages on each of the platforms.  You can, of course, use a git URL in either case, but ideally you want to use managed packages, not just pull from a git repo.  So it would be worth someone going through the packages that are already being used, and checking availability.

Another thing, which I'll admit is more just a personal preference, is that it keeps browser-side code away from tooling such as `gulp` and its packages.

Personally I lean more towards bower for browser-side code, mainly for the flattened dependencies.  However in the case of CA, optimisation is not (in general) such an issue, and if everyone on the project is more familiar with npm, and the packages are available, then I see no reason not to go with it.

Feel free to copy-paste this somewhere else with my name on it, or let me know and I will do it, if there is a more appropriate place to discuss this than the bottom of a semi-related PR ;-)

collectiveaccess:
Well then it's Bower then :-)

gautiermichelin:
:-) 👍

PR #276, opened 2016-12-12
PROV-1883 List Items missing in hierarchical or type-ahead look up

brunetton:
This commit doesn't concentrate to the origin of missing "name" in received json items, it's only a workaround.

### DIFF
--- a/assets/ca/ca.hierbrowser.js
+++ b/assets/ca/ca.hierbrowser.js
@@ -400,7 +400,13 @@ var caUI = caUI || {};
 					for(var i in data['_sortOrder']) {
 						var item = data[data['_sortOrder'][i]];
 						if (!item) { continue; }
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Thanks for this. I'll see if I can come up with a more permanent fix shortly.

brunetton:
Speaking of hierarchical widget; I have another bug that prevent from Ajax loading more than two times when scrolling to the end of list on widget. I tried on a 500 elems list, changing the max number of loaded elems to 50, and I can't get more than 100 elems displayed.
Am I the only one to have this problem ? I tried with Chromium and FIrefox browsers, same effect.
If you noticed that too; then I'll fill a bug.

collectiveaccess:
I haven't noticed it but it doesn't mean there's not a problem. Most hierarchies stay under 500 so I wouldn't be surprised if no one has complained.

PR #280, opened 2016-12-23
PROV-1960 Add notice when list element not found and not created due …

brunetton:
Minor change to close the subject: I need to know while importing when an item hasn't been found. In my case, this shouldn't happen; I need a warning, go to source file and fix the problem.
I also put this portion of code after the checking of list existence.


collectiveaccess:
Thanks!

brunetton:
You're welcome !

PR #281, opened 2016-12-27
Prov 1962: make non valid json in option a fatal error for mapping import + fix: no warnings while importing mapping from console

brunetton:
Two fixes here:
  * [Prov 1962](http://clangers.collectiveaccess.org/jira/browse/PROV-1962) make non valid json in option a fatal error for mapping import
  * *NOJIRA* fix no display of warnings while importing a mapping from command line with `CLIUtils.php`

### DIFF
--- a/app/lib/ca/Utils/CLIUtils.php
+++ b/app/lib/ca/Utils/CLIUtils.php
@@ -1154,8 +1154,8 @@
 				CLIUtils::addError(_t("Could not import '%1': %2", $vs_file_path, join("; ", $va_errors)));
 				return false;
 			} else {
-				if(is_array($va_errors) && (sizeof($va_errors)>0)) {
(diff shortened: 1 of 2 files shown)


brunetton:
I never know if I have to request to merge to master or develop branch. Say me if you prefer PR-ing to develop.
Thanks

collectiveaccess:
Actually can you PR to develop? That's going to be the release in just a few days. Thanks.

brunetton:
Ok !
Why don't we use the wiki here for putting some basic dev stuff ? Like branches use, variable naming conventions (for example, in $pa_source_data, what "pa" stands for ? Pointer to Array ?)

collectiveaccess:
We already have the other wiki (MediaWiki) and a new Sphinx-based documentation set. I fear another documentation source is two too many. That said, my hope is to be rid of the MediaWiki soon, so I'd be open to using the GitHub wiki for just only internals documentation.

PR #286, opened 2017-02-10
PROV-2006

ammopt:
This fix updates the name argument used in method getSourceListAsHTMLFormElement with the actual name of the form element (placement_code+form_prefix+element_name), instead of just "source_id".


ammopt:
Added a second commit for PROV-2002.

PR #290, opened 2017-03-23
add table alias to type limit part of sql

naeemmuhammad:
The sql query that is being used to remove relationships, when relationship type_ids are provided, needs to be contextualized by adding alias (of the relationship table) to the type_id column. All columns used in this query are being prefix with corresponding alias, except type_id. Because type_id belongs to the relationship table therefore its alias ‘r’ (defined in the query) needs to be used.
The above mentioned query is constructed and executed in function ‘removeRelationships’. This function is used while batch editing relationships. We had some trouble in removing and replacing relationships in batch editing. Adding alias to type_id solves the problem.


collectiveaccess:
Thanks for this. We need this applied to master-fix, not master though. I have gone ahead and manually applied the patch to both master-fix and master.

PR #291, opened 2017-03-24
put bundle editor popup screen atop

naeemmuhammad:
Sometimes display bundle editor popup screen cannot be closed because bottom of the screen (that contains the control to close it) is underneath the next parent screen component (see an example screen below). Only way to close this popup screen is to refresh the page, which is quite annoying. This problem can be solved by putting popup screen at the top of the components stack. Fix in this pull request does that.

Screenshot without the fix:

Screenshot with the fix:


collectiveaccess:
Thanks for this! 
Merged this to master-fix. In the future can you do pull requests on master-fix instead?

PR #299, opened 2017-08-03
Fix search parsing infinite loop / memory-overflow error.

leftclickben:
When attempting to determine whether a given search includes Sets, it is parsing the search but when there are parentheses `()` present, it runs into an infinite loop.

This resolves the infinite loop by removing the parentheses from the search text.  This doesn't actually affect whether or not the search is for sets.

### DIFF
--- a/app/helpers/searchHelpers.php
+++ b/app/helpers/searchHelpers.php
@@ -1617,7 +1617,7 @@
 		$o_query_parser->setEncoding($o_config->get('character_set'));
 		$o_query_parser->setDefaultOperator(LuceneSyntaxParser::B_AND);
 		
-		$ps_search = preg_replace('![\']+!', '', $ps_search);
+		$ps_search = preg_replace('![\'()]+!', '', $ps_search);
 		try {
 			$o_parsed_query = $o_query_parser->parse($ps_search, $vs_char_set);
 		} catch (Exception $e) {


leftclickben:
@collectiveaccess I think this is being pushed to the right branch.  If not, let me know and I'll change the PR.

collectiveaccess:
What code did you see this behavior in? It should have been fixed with this commit two months ago:
https://github.com/collectiveaccess/providence/commit/789944353dae6b7c7c2da51758b3e4401fb615d9

That fix does work for me with searches that include parens.

leftclickben:
Thanks @collectiveaccess 

The bug was introduced in cdd4ea06a68f74c4f350a732795553d34056e6a1, which is present in `master` and all the `1.7` tags, but not in `master-fix`.

Obviously I should have looked in `develop` and I would have found your fix, but I guess I thought it would only contain new development work, and that bugfixes should be on `master-fix`..?  

Actually, `master` and `master-fix` are quite divergent and it is not possible to use `git bisect` directly between them.  I used a common ancestor to find the commit where the bug was introduced, then I was able to come up with my somewhat hacky fix.

Anyway, I'll close this because assuming your fix works, we will use that instead.

inquisite:
master-fix hasn't been maintained – as you have observed – since the 1.7 release. We'll put out a 1.7.4 release in the next day or so, and bring it up to date. The idea is actually to have much less time between minor releases going forward, which obviates the need for a master-fix branch to some extent.

PR #302, opened 2017-09-07
NOJIRA Make it possible to ignore errors when importing.

kehh:
* This allows for debugging import mappings that would otherwise stop on first error.

### DIFF
--- a/app/models/ca_data_importers.php
+++ b/app/models/ca_data_importers.php
@@ -1540,20 +1540,7 @@ class ca_data_importers extends BundlableLabelableBaseModelWithAttributes {
 			// Get data for current row
 			//
 			$va_row = array_replace($o_reader->getRow(), $va_environment);
-			
-			// replace values (EXPERIMENTAL)
-			$va_row_with_replacements = $va_row;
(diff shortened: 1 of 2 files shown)


app/models/ca_data_importers.php
```diff
@@ -1540,20 +1540,7 @@ static public function importDataFromSource($ps_source, $ps_mapping, $pa_options
 			// Get data for current row
 			//
 			$va_row = array_replace($o_reader->getRow(), $va_environment);
-			
-			// replace values (EXPERIMENTAL)
-			$va_row_with_replacements = $va_row;
-			foreach($va_mapping_items as $vn_item_id => $va_item) {
-				$vs_key = $va_item['source'];
```
collectiveaccess: What is the reason for removing this?

kehh:
Bad cherry picking. Will put up a new one later.

kehh:
@collectiveaccess I have restored the bits that got removed in d04d3f4. Let me know if this is sufficient. We've been using this quite a bit lately for CSIRO and it's really useful to debug import mappings and run test imports.

PR #303, opened 2017-09-13
Support SSL by detection rather than global.conf

kehh:
`global.conf` variables cannot be overridden, so this is in lieu of changing the variable locally in that file. This way CA detects if it's using SSL from the server.


kehh:
The failing tests do not related to this current change.

collectiveaccess:
Yes, that's just the Getty AAT service acting up.

PR #308, opened 2017-11-07
Update processTaskQueue.php

starfruitsolutions:
I was having trouble with this script failing to select the correct path for caUtils depending on the directory it was executed from. This method of acquiring the path is more succinct and should work in all cases.

### DIFF
--- a/support/utils/processTaskQueue.php
+++ b/support/utils/processTaskQueue.php
@@ -49,10 +49,7 @@
 	$_SERVER['argc'] = $argc;
 	
 	ob_start();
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Thanks for this. The processTaskQueue.php script is a legacy script we keep around for compatibility with old installs. Have you tried running the task queue using support/bin/caUtils?

starfruitsolutions:
Not directly. The providence install documentation instructs to run processTaskQueue.php. When it failed to find caUtils, I opened the file and noticed a fix was simple.

collectiveaccess:
ok thanks for the fix!

PR #313, opened 2017-11-23
NOJIRA Allow parents to specify options in caProcessRefineryParents

kehh:
* matchOn and ignoreParent can now be set for each parent.
* If we're ignoring parents then actually ignore them.
* getListItemID takes `idno` as the second parameter, so pass that.
* Pass correct preferred labels in with list item.
* Fix missing array paremeter to `array_map()` call.

### DIFF
--- a/app/helpers/importHelpers.php
+++ b/app/helpers/importHelpers.php
@@ -1,6 +1,6 @@
 <?php
 /** ---------------------------------------------------------------------
- * app/helpers/importHelpers.php : 
+ * app/helpers/importHelpers.php :
  * ----------------------------------------------------------------------
  * CollectiveAccess
  * Open-source collections management software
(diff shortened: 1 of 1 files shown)


kehh:
This allows us to import a hierarchical table where you specify a parent id in a specific column.

An example listItemHieararchyBuilder configuration:
```json
{
  "parents": [
    {
      "idno": "ento-^intParentID",
      "name": "^parentFullName",
      "ignoreParent": true,
      "matchOn": ["idno"],
      "type": "entoTaxon",
      "attributes": {
        "ca_list_items.is_enabled": true
      }
    }
  ],
  "list": "taxonomy"
}
```
This allows us to set `ignoreParent` and `matchOn` for each individual parent.

app/helpers/importHelpers.php
```diff
@@ -284,38 +286,38 @@ function caProcessRefineryAttributes($pa_attributes, $pa_source_data, $pa_item,
...
 							if (is_null($vn_num_repeats)) { $vn_num_repeats = sizeof($va_vals); }
-							
+
 							$vn_c = 0;
 							foreach($va_vals as $vn_x => $va_v) {
 								if (!$va_v || (!is_array($va_v) && !trim($va_v))) { continue; }
 								if ($vs_prefix && is_array($va_v)) {
-									$va_v = array_map(function($v) use ($vs_prefix) { return $vs_prefix.$v; });
-									
+									$va_v = array_map(function($v) use ($vs_prefix) { return $vs_prefix.$v; }, $va_v);
```
kehh: No array was being passed to array_map() here.

kehh:
The failed test appears unrelated.

PR #314, opened 2017-11-24
Refine list export options when exporting configuration.

kehh:
* add `configuration_export_exclude_list_items_larger_than` and `configuration_export_exclude_list_items` options.
* Whitespace trimmed from end of lines by PHPStorm/.editorconfig
NOJIRA

PR #315, opened 2017-11-30
Refine list export options when exporting configuration.

kehh:
* add `configuration_export_exclude_list_items_larger_than` and `configuration_export_exclude_list_items` options.
Implements #314  against develop branch.


yelto:
Could the names lead to confusion given the existing parameters? Something along these lines be might more self-explanatory:
configuration_export_exclude_list_items_from_these_lists
configuration_export_exclude_list_items_from_lists_larger_than

PR #316, opened 2017-11-30
Allow parents to specify options in caProcessRefineryParents

kehh:
* matchOn and ignoreParent can now be set for each parent.
* If we're ignoring parents then actually ignore them.
* getListItemID takes `idno` as the second parameter, so pass that.
* Pass correct preferred labels in with list item.
* Fix missing array paremeter to `array_map()` call.
This implements #313 against develop branch.

### DIFF
--- a/app/helpers/importHelpers.php
+++ b/app/helpers/importHelpers.php
@@ -1,6 +1,6 @@
 <?php
 /** ---------------------------------------------------------------------
- * app/helpers/importHelpers.php : 
+ * app/helpers/importHelpers.php :
  * ----------------------------------------------------------------------
  * CollectiveAccess
  * Open-source collections management software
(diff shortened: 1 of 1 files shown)


kehh:
This allows us to import a hierarchical table where you specify a parent id in a specific column.

An example listItemHieararchyBuilder configuration:
```json
{
  "parents": [
    {
      "idno": "ento-^intParentID",
      "name": "^parentFullName",
      "ignoreParent": true,
      "matchOn": ["idno"],
      "type": "entoTaxon",
      "attributes": {
        "ca_list_items.is_enabled": true
      }
    }
  ],
  "list": "taxonomy"
}
```
This allows us to set `ignoreParent` and `matchOn` for each individual parent.

PR #317, opened 2018-01-02
Prettify "counts"-display by adding a non-breaking space.

felixmz:
Makes a real difference in the german localization, where the words need a little more characters...


collectiveaccess:
Thanks!

PR #327, opened 2018-05-04
Add description to list item labels.

kehh:
* the configuration exporter now exports the description.


kehh:
(This is because df37de2804934103ffa327d3e826b430b91f7fc4 now exports list item labels).

PR #333, opened 2018-07-31
Adding new customizable InformationService for SPARQL queries, Wikida…

chirale:
…ta as default values


collectiveaccess:
Thanks!

PR #341, opened 2018-09-25
Other users public sets

naeemmuhammad:
This implementation adds other users' public user sets (access value 'accessible to public') to the 'Search by set' drop down list. This drop down list is in the search interface. Currently this list only contains user's own public and private sets, whereas this implementation brings in public user sets of other users.
 
image


collectiveaccess:
Why would this be desirable on the back-end?

naeemmuhammad:
What is public access modifier for user sets? It seems they should be available to other users. And if other users want to access them they should be available for search in search screen.

collectiveaccess:
Updated this in PR-605 and merged. Thanks for submitting this and apologies for letting it lie for 2 years :-/

PR #344, opened 2019-03-11
NOJIRA : fix dimensions length value containing value (decimal separa…

gautiermichelin:
NOJIRA : fix dimensions length value containing value (decimal separator for french, spanish…)

Capture d’écran 2019-03-11 à 16 57 17

Decimal delimiter is actually cleaned up, keeping only the dot as a decimal delimiter. 

(all of the green countries are concerned https://en.wikipedia.org/wiki/Decimal_separator)


gautiermichelin:
Seem to be some differences between the versions. I close this PR.

PR #348, opened 2019-06-01
Update Vagrantfile to use php7.2 on ubuntu xenial

tasinttttttt:
The current 1.7.8 version of providence doesn't work with the provided Vagrantfile.
So I've updated it to use php7.2 with the required — at install at least — dependencies.

The current Vagrantfile with php5.5 fails with a 500 error (I suspect something involving vendors...).
Also tried with php7.3 and that fails at the installation page 2 (only creates some tables and then freezes).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,7 +3,7 @@
 
 Vagrant.configure(2) do |config|
 
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "ubuntu/xenial64"
 
 
   # Create a forwarded port mapping which allows access to a specific port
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Thanks for catching this! We haven't tested the vagrant file in quite some time. With regards to PHP 7.3, we have not tested CA with it yet, but are starting that process. The next version will be compatible.

Vagrantfile
```diff
@@ -15,61 +15,68 @@ Vagrant.configure(2) do |config|
...
 
     echo "mysql-server mysql-server/root_password password root" | sudo debconf-set-selections
     echo "mysql-server mysql-server/root_password_again password root" | sudo debconf-set-selections
     apt-get -y install mysql-client mysql-server
 
-    apt-get -q -y -o Dpkg::Options::=--force-confold install apache2
-    apt-get -q -y -o Dpkg::Options::=--force-confold install php5 libapache2-mod-php5 php5-cli
-    apt-get -q -y -o Dpkg::Options::=--force-confold install php5-curl php5-mysqlnd php5-json php5-gd php5-imap php5-mcrypt
+    apt-get -q -y -o Dpkg::Options::=--force-confold install curl apache2
+    apt-get -q -y -o Dpkg::Options::=--force-confold install php7.2 libapache2-mod-php7.2 p7.2-curl php7.2-mysql
```
tasinttttttt: Typo here !
```
+ php7.2-curl
- p7.2-curl
```

PR #353, opened 2019-08-29
NOJIRA Don't remove empty values from arrays in caSanitizeArray

kehh:
* The method description says it's designed to strip weird noncharacter data.
* This change only removes values if they are actually set.
* Means that SimpleAPI can return blank values in a deterministic way.


collectiveaccess:
Thanks

PR #355, opened 2019-10-01
NOJIRA Fix full text indexing of entity labels and other fields.

kehh:
* Don't index as idno by default - in_array required strict checking.

### DIFF
--- a/app/lib/Plugins/SearchEngine/ElasticSearch/FieldTypes/Intrinsic.php
+++ b/app/lib/Plugins/SearchEngine/ElasticSearch/FieldTypes/Intrinsic.php
@@ -118,7 +118,7 @@ class Intrinsic extends FieldType {
 			$this->getTableName() . '/' . $this->getFieldName() => $pm_content
 		);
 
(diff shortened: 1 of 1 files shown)


app/lib/Plugins/SearchEngine/ElasticSearch/FieldTypes/Intrinsic.php
```diff
@@ -118,7 +118,7 @@ public function getIndexingFragment($pm_content, $pa_options) {
 			$this->getTableName() . '/' . $this->getFieldName() => $pm_content
 		);
 
-		if($t_instance->getProperty('ID_NUMBERING_ID_FIELD') == $this->getFieldName() || (is_array($pa_options) && in_array('INDEX_AS_IDNO', $pa_options))) {
+		if($t_instance->getProperty('ID_NUMBERING_ID_FIELD') == $this->getFieldName() || (is_array($pa_options) && in_array('INDEX_AS_IDNO', $pa_options, true))) {
```
kehh: Does this make sense? The config is parsed like as follows:
```
array:3 [
  "BOOST" => "100"
  0 => "DONT_TOKENIZE"
  1 => "TOKENIZE"
]
```
Without this fix entity labels are exploded into their individual parts which makes it impossible to match exact strings with spaces.

So the search clause ca_entity_labels:"Department of Obscure Things" would not match a record where the associated entity is "The Department of Obscure Things".

PR #360, opened 2020-01-01
Add a missing PHP open tag

miceno:
The template was broken due to a missing PHP open tag.


miceno:
Already solved in `develop` branch.

PR #361, opened 2020-01-01
PROV-2724 Propagate dryRun to data importer

miceno:
It solves https://collectiveaccess.atlassian.net/browse/PROV-856 and https://collectiveaccess.atlassian.net/browse/PROV-2724.

### DIFF
--- a/app/lib/Utils/CLIUtils/ImportExport.php
+++ b/app/lib/Utils/CLIUtils/ImportExport.php
@@ -455,8 +455,9 @@
 
 			$vb_direct = (bool)$po_opts->getOption('direct');
 			$vb_no_search_indexing = (bool)$po_opts->getOption('no-search-indexing');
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Thanks for this. Have you tested dry run in the user interface (not command line)? We've seen issues with dry run being run from there where it doesn't work, even though it's being set properly. In those cases it appears to be an issue with transaction handling.

miceno:
I tested it on the command line and on the UI, but not thoroughly, I tested it for my main use case, with the following command line (same on the UI):

`php7.2 bin/caUtils import-data --source ~/incoming/input-data.xlsx  --mapping import-mapping -f XLSX -d DEBUG --dryrun -t import-set`

collectiveaccess:
What version of MySQL are you running?

miceno:
MySQL 5.5.60

PR #371, opened 2020-03-23
NOJIRA Add a special source for current date.

miceno:
I was looking for the way to set one of the field to the current import date, so you can trace it later, when the record was imported.

The creation date might be also that field, but it is an intrinsic field, I would like to add it to any date field in the profile.

I came across a patch for it, very simple ;)


collectiveaccess:
Looks good!

PR #372, opened 2020-03-24
Allow logging on reprocess media

miceno:
I had to verify that `reprocess-media` command line tool was running properly and I made this patch.

I have mimic the behaviour of other commands and also took the opportunity to refactor code to create loggers, so there is also the opportunity to refactor other code pieces in the future, and avoid boilerplate code (see new function `caGetLogger`).

Hope it will be useful for other people too!


collectiveaccess:
Oh excellent! This was on my todo list :-)

PR #380, opened 2020-04-02
Allow caching test env for faster test execution

miceno:
Allow caching `composer` and database for faster env setup, test time reduction of about 30-50%, with same reliability.

It uses Travis-CI caching facility for caching:
- `vendor` folder. It will copy `composer.lock` file to `vendor` folder too, so restoring the configuration is extremely fast.
- profile database dump. It will dump database after profile installation. It will load it again on the next execution. It will speed database creation (creating relationships and other tables are really slow). If you would like to skip profile database caching, set env variable `SKIP_CACHED_PROFILE` to any non-empty value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ dist: bionic
 addons:
   apt:
     packages:
-    - mysql-server-5.7
-    - mysql-client-core-5.7
-    - mysql-client-5.7
+      - mysql-server-5.7
+      - mysql-client-core-5.7
+      - mysql-client-5.7
 
 git:
   depth: 5
@@ -21,15 +21,18 @@ cache:
   - directories:
       # Cache vendor folder will speed up environment setup
       - vendor
+      - tests/mysql_profile
 
 services:
   - mysql
 
 env:
-  # Here we list the profiles that we want to test, each one will be tested, per version of PHP, in parallel
-  - PROFILE=testing
+# Here we list the profiles that we want to test, each one will be tested, per version of PHP, in parallel
+#- PROFILE=default
 
 before_install:
+  # execute all of the commands which need to be executed
+  # before installing dependencies
   # Remove xdebug to increase speed
   - phpenv config-rm xdebug.ini
   # Set up database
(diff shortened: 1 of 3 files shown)


miceno:
I suggest running the build twice in Travis-CI in order to validate it, since the first execution will create the cache, and the second will use it.

miceno:
Main reason behind this PR is that running on Travis-CI for php 7.4, profile install breaks due to a MySQL timeout (see https://travis-ci.org/github/collectiveaccess/providence/jobs/670287309).

This PR allows for a workaround. You should commit a profile database dump from a local setup. Create the testing database using script `database_setup.sh`, and then create the testing profile with `profile_setup.sh`. Commit file `tests/mysql_profile/testing.sql` and rerun the PR build.

Once the cache is created for php 7.4, commit again removing profile dump  `tests/mysql_profile/testing.sql`.

collectiveaccess:
Isn't the real problem to solve here why there a lock timeout in the first place?

miceno:
Sure it is!

I tried it several different ways, I tested in my local environment and I couldn't reproduce it. I think you may need to report the issue too, I have already done it here https://travis-ci.community/t/broken-builds-on-php-7-4/7871, but no success as of today.

So I tried different things, and some of them are shown on this PR:

- modified MySQL server configuration to rise the timeout. I rised it to 300 seconds and it didn't work, but the code for adding MySQL server configuration works, so I added it to the PR.
- restoring a full dump. In the meantime, I also found a way to cache them, so it is a valid strategy for non-breaking builds.

In fact, the code in this PR might also work even if there is not such issue on Travis-CI, and I thought it might be relevant for the project. 

Nevertheless, I think that restoring a database dump is to some extent a bit aggressive :( but I didn't find other way.

Do you suggest other approaches?

collectiveaccess:
Let me look into it. It has been very busy here and I haven't had time to investigate the erroring. Thank you for working on it!

miceno:
Hi again!

I tried to debug Travis-CI, and I finally manage to manually run the install successfully on PHP 7.4, but I don't know how I solved it. I just added a print statement with the sql to execute before it is executed, here `/home/travis/build/miceno/providence/app/lib/Db/DbStatement.php(151)`, and it worked. Then I removed the print statement, and it worked again :(

I think the issue is related to the way BLOBs on the changelog are created. I think there is something that changed on the way binary strings are manage from php 7.3 and php 7.4. What do you think?

collectiveaccess:
No idea. I'll try to reproduce this on my test machine.

PR #381, opened 2020-04-04
Fix minor errors and add todo for unclear issues.

miceno:
I run inspection code on folder `app/helpers` using PHPStorm and it revealed several warnings.

I tried to solve most of them, but some others are not clear to me, if there are bugs or whether the flow will finally allow variables to be available.

When it was not clear, I add a `TODO` comment, so the maintainer may advice on how to proceed.

On those that I tried to solve, I may made a mistake, I tried my best to understand the flow but nobody like the author to comment on this.


collectiveaccess:
We can't merge this as-is. It really should be multiple commits addressing specific issues in specific locations. Some of these are good, and some will break things.

I'll go through these one by one and resolve them. 

Thanks for doing this. It's super-helpful.

miceno:
Closing the PR.

collectiveaccess:
Added new dev/code-fix branch with suggested changes reviews and applied. Will test and merge this week.

miceno:
Thank you!!! 👍

PR #385, opened 2020-04-05
Allow additional arguments for ImageMagick external applications

miceno:
Define a configuration item in `external_applications.conf`, like this: `imagemagick_<COMMAND_NAME>_args` per command.

For example, for `convert` command, use 

```
imagemagick_convert_args = -limit thread 1
```

As of now it only allows arguments on the ImageMagick media class. It is pending a refactor of TilePicParser class to use this feature.

With a bit of refactor, new class `CommandConfiguration` can be promoted to `BaseMediaPlugin` class, and the feature will be available to subclasses.

### DIFF
--- a/app/conf/app.conf
+++ b/app/conf/app.conf
@@ -1618,6 +1618,9 @@ always_show_counts_for_relationship_bundles_in_editor = 0
 #
 ca_sets_default_type = user
 
+# Allow multiple instances of the same item in a set?
+allow_duplicate_items_in_sets = 0
+
 # -----------------------------------
 # Timecode output
 # -----------------------------------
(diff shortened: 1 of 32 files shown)


miceno:
I am afraid I made something wrong rebasing from upstream develop branch...

Do you think I should create a new PR cleaning the code?

collectiveaccess:
What is wrong?

miceno:
There are more changes in the PR than those from my initial branch, see changed files. I think I rebased using upstream develop and it included more commits.

collectiveaccess:
Yes, please resubmit. Thanks!

PR #387, opened 2020-04-06
Changed Time Expression Parser for locale pt_PT

j-n-c:
Changed Time Expression Parser for locale pt_PT


collectiveaccess:
Thank you!

PR #390, opened 2020-04-11
Show max upload limit on every file upload.

miceno:
Additionally refactor code and move formatting function to utility helpers.

### DIFF
--- a/app/helpers/utilityHelpers.php
+++ b/app/helpers/utilityHelpers.php
@@ -4274,3 +4274,19 @@ function caFileIsIncludable($ps_file) {
 			);
 	}
 	# ----------------------------------------
+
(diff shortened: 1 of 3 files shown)


collectiveaccess:
This is a lot of text to add... how does it look when you try it?

miceno:
Long, but I think a user needs the long tail to understand it. I hereby attach a screenshot.

image

miceno:
I had to upload a 180Mb file and the configuration in php.ini was below that, so the request uploaded the file, but it silently complained later, with a CSRF Invalid Token error.

I would not have uploaded the file is I had known in advance the size limit, so I thought adding the text with the limits will help other users too.

collectiveaccess:
It's way too long, but let's merge it and see how we can make it shorter later.

PR #391, opened 2020-04-12
BUG Media viewer does not use image tile size.

miceno:
I configured media_processing rules to allow a tile size of 512px and then images are shown with disordered tiles.

I traced the error to the way the tile viewer is configured in `htmlFormHelpers.php`, and there is a hardcoded value of `256` for the tilesize of the viewer.

I also added some tests.


collectiveaccess:
It's way too long, but let's merge it and see how we can make it shorter later.

PR #393, opened 2020-04-18
FIX Duplicate entry for `log` and `limit-log-to`.

miceno:
Both arguments was sharing `l` (lowercase L).


miceno:
Hi,

I realized @collectiveaccess  made a commit that conflicted with this PR, to set the option to `-g` and not `-L`.

I will close this PR then.

PR #396, opened 2020-04-20
FIX Remove duplicate for `lifespan` intrinsci attribute.

miceno:
There is an intrinsic attribute `ca_entities.lifespan` as per https://github.com/collectiveaccess/providence/blob/9f4a96639f80da0e31e9acc9b76bb8935f8c36b6/app/models/ca_entities.php#L109-L117

And there is an additional `lifespan` metadata attribute on `default.xml` profile. 

This PR removes the second one.


miceno:
Thank you!!!

PR #399, opened 2020-04-22
NOJIRA Add configuration to allow setting curl connection timeout.

miceno:
While running unit test, AAT test broke due to a connection timeout. It was hardcoded to 2 seconds, but the endpoint sometimes took more than 5 seconds.

I added some code to allow configuration of curl options. I think the code can be moved to a helper, so other Information Services will reuse it, but this is the first step.

Configuration files for server and testing are also updated.


miceno:
Super-fast! Thank you!

Issue #400, opened 2020-04-23
Too many failures testing on local but not in TravisCI


miceno:
Hi 

I am testing the latest latest develop branch on my local computer (macosx) using php 7.3.13, and it breaks the tests :( The report says there 149 failed tests.

The point is that I had a look at TravisCI and the builds are ok, so I thought it was something on my environment.

I debug a bit and I discovered a couple of things that look like the cause of some of the failed tests, but what it is really puzzling me is why those tests didn't fail in TravisCI :(

In fact, this is the way I found the timeout issue on the AAT curl requests, because the test failed locally.

For example, I found that `ConfigurationUpdateTest::testDeleteList` failed because there was a missing DB transaction commit here before `continue` statement: 
https://github.com/collectiveaccess/providence/blob/d99ab18a516c87c486314cf0f7a2425fd784b0ba/install/inc/Installer.php#L590-L593

But I think that failure should also show up on TravisCI, but it didn't. Weird!

I created a clean environment cloning develop branch on a new folder, and run it again, and same failures...

I run it also on PHP 7.2.26, and it also raised the around 140 failures.

I am attaching the test report:
[Providence develop - ALL_tests.html.zip](https://github.com/collectiveaccess/providence/files/4524488/Providence.develop.-.ALL_tests.html.zip)

I now I can run tests on TravisCI, but I prefer running them locally, it is faster and less time consuming.

Any suggestion on how to deal with the problem?

collectiveaccess:
Did you run this on a clean install of the test profile?

miceno:
Yes, I did. I remove the database, I create it again, and I also clean the app/tmp/collectiveaccess cache folder

collectiveaccess:
And you're sure it's actually using that database? Many of the errors imply you may not be. Also, how are you running these tests?

miceno:
Yes, it is using the testing database. I am debugging it and the testing database changes.

I am running the tests from PHPStorm, here is the command line shown:

/usr/local/Cellar/php@7.3/7.3.13/bin/php providence/vendor/phpunit/phpunit/phpunit --configuration providence/tests/phpunit.xml --teamcity

collectiveaccess:
What version of PHPUnit do you have installed? I'm running under 7.3 as well...

miceno:
PHPUnit 8.5.0 by Sebastian Bergmann and contributors.

miceno:
Let me try it from the command line, without  PHPStorm IDE... I will report back...

[...]

PR #402, opened 2020-04-24
Refactor common code in Getty Information Service code.

miceno:
Move common code to `lookup` member in parent class so the lookup code is shared by different subclasses.

It is related to issue #400


miceno:
Thank you!

PR #403, opened 2020-04-24
Force loading a clean setup from the dist file.

miceno:
Bootstraping file for tests at `setup-tests.php` loads `setup.php` file.

In case you are developing locally, it will use your local configuration too, so tests might break, for instance, if you setup a timezone in Europe, or any other local configuration.

A workaround is already in place in `.travis.setup.php` file: it loads vanilla setup from `setup.php-dist`.

Related to #400.


miceno:
Thank you!

PR #404, opened 2020-04-24
Mark tests as incomplete when web service lookup fails.

miceno:
Related to #400.

A Web service is an external dependency. In case it fails I suggest marking the test as incomplete.


miceno:
Thank you!

PR #414, opened 2020-05-03
Remove short alias on global options to avoid conflicts on subcommands.

miceno:
I realized that short alias for global options may conflict with sub-commands options, for example, `import-media` source option also uses `-s` as a short alias (a PR will follow also on `import-media` params later).

In order to avoid the conflict, I removed sort aliases on global options.

In the process I realized:

* the right syntax for a long-only option is `longoption[-=]s`, so I also updated the code accordingly.
* former `_formatCmdOptions` function will best fit into a helper file. I refactored it a bit, using code borrowed from Zend GetOpt implementation.
* there were no tests on the former `_formatCmdOptions`, so I add them in this PR.

### DIFF
--- a/app/helpers/CLIHelpers.php
+++ b/app/helpers/CLIHelpers.php
@@ -142,3 +142,24 @@ function caSetupCLIScript($pa_additional_parameters) {
 	return $o_opts;
 }
 # ---------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


collectiveaccess:
Looks good. Thanks!

PR #420, opened 2020-05-10
FIX Fuzzy search caused infinity recursion

miceno:
While running a query like "centro~" the server crashed with an HTTP 500 Error.

Tracing the cause, it was due to an infinity recursion on `caSearchIsForSets`, because there was no stopping condition when subqueries were fuzzy.

Added some tests too.


collectiveaccess:
Thanks for this. I don't think anyone has every thought to use the fuzzy search operator. It won't actually do anything special with SqlSearch.

We'll review this all and merge soon.

miceno:
Sure! 

I was giving it a try when HTTP 500 arised :(

collectiveaccess:
Keep it coming please :-) This is turning into a full time job for you. For this I apologize.

PR #428, opened 2020-05-22
Compact time progress format.

miceno:
Use a more compact representation of the progress ETC and ellapsed times.


collectiveaccess:
Good idea. Thanks!

miceno:
Thank you!

PR #433, opened 2020-05-26
FIX undefined variable in refactored code

miceno:
Fix an undefined variable on refactored code.


SethKaufman:
Thanks for catching that

miceno:
You're welcome.

I think this is part of the reason the new feature didn't work :( 

So I think we should carefully think about the development cycle, see #434 so we can elaborate it a little bit more.

PR #474, opened 2020-06-15
User Generated Recommendations

AlanBridgeman:
Okay, so, its been a while since I've contributed to something on GitHub so I hope I've done this correctly. Through my changes to the branch of my fork of providence I've added support for UserRecommendations by introducing two new hooks and then a plugin. Search still doesn't seem to be working entirely correctly but. the rest seemed to be. I'm contributing this in the hope someone might find this useful if only as a starting off point.


collectiveaccess:
Thank you for the contribution! Can you explain a bit what these code change are supposed to accomplish? What exactly is a "UserRecommendation"?

AlanBridgeman:
Sure, my apologies for the lack of clarity initially. User Recommendations loosely based on User Generated Content controls already existing in CollectiveActive are to give members of a community an ability to effect the organization and structure of the system that is to provide users/viewers of the system the ability to recommend collections that an object can be a part of. This could be particularly useful if you had a group of artwork or records with limited metadata or information about that artwork or records and you wanted to "crowdsource" that work allowing users/viewers of the system to provide you informations by recommending that a piece should be part of a given collection. Note, that one very important aspect is administrative oversight that is where, at least in my understanding, currently you can give people access to make their own collections this plugin in contrast would allow the administrators of the artwork or record group to say if something should be a part of the recommended collection. This also allows if you work with sensitive artwork or records the control on privacy if something should be denoted as being privately part of a collection versus publicly part of a collection.

I should note that this is only the "back-end" Providence first "draft" implementation. However, I felt it would be a worthwhile contribution mostly due to the needed modifications to specific core files. Most notably, my addition of a hookNewModel within the DataModel class which in my hope would allow Plugin developers the ability to create and contribute their own models if needed to the Providence system. Thus hopefully spurring even further development of Providence and CollectiveAccess as a whole.

collectiveaccess:
We cannot merge this as-is, but will investigate incorporating this code into CA to implement similar functionality in the future. Thank you for contributing!

PR #489, opened 2020-06-24
change guzzle query request array

the666bbq:
Hi,
the query parameter for this guzzle request to viaf was set in a separate array and should be nested in the options array. 
The old code resulted in a 500 error at VIAF, now results in the resultset by VIAF

### DIFF
--- a/app/lib/Plugins/InformationService/VIAF.php
+++ b/app/lib/Plugins/InformationService/VIAF.php
@@ -80,7 +80,7 @@ class WLPlugInformationServiceVIAF extends BaseInformationServicePlugin implemen
             'headers' => [
                 'Accept' => 'application/json'
             ],
-            ['query' => "'".$ps_search."'"]
+            'query' => ['query'=>$ps_search]
         ]);
         #$vo_request->setHeader('Accept', 'application/json');
         #$vo_request->getQuery()->add('query', "'".$ps_search."'");


collectiveaccess:
Thanks for this! 

We have a patched version of VIAF in the development branch that also fixes this. I'll compare the two; you may want to give it a try too.

the666bbq:
I see that the develop code goes the queryparam way, and has extra arguments for the search. guzzle maybe has some advantages being more clean and handling the url_encoding already ? Don't know the use-cases of Viaf enough to judge the extras.

slightly unrelated-related, is there a documentation on the CA workflow to get the latest safe version of the codebase. We base our installation on the latest (1.7.8) but there are many goodies on master-fix and develop branches clearly (maybe other branches too?). Or is there a timeline somewhere for new minor of major version that have the goodies :)

collectiveaccess:
The extras matter in terms of search quality, or so I'm told. I'm not a VIAF user either.

The develop branch is the "safe" up-to-date branch. Everything else under dev/* is unstable. Everything under host/* is maintained for stability of various hosting environments.

I keep on saying the next release will be in a week, and then something happens. But yeah, develop will turn into a release soon... let's say within a month.

the666bbq:
That's very good news. I did find a 'recent' documentation last week with 1.8 mentioned 📦 

I cannot get the develop version of VIAF working though. The URL that is constructed is 👎 
[(http://www.viaf.org/viaf/AutoSuggest?maximumRecords=100&httpAccept=application/json&query=cql.any+all+%22mozart%22)]

but the CQL viaf query syntax apparently goes through the "search" endpoint, not the "AutoSuggest" (I read here [http://www.viaf.org/viaf/search] and here [https://viapy.readthedocs.io/en/latest/codedocs.html] and indeed [http://www.viaf.org/viaf/search?maximumRecords=100&httpAccept=application/json&query=cql.any+all+%22mozart%22] does work but give completely different output, and doesn't contain the 'result' variable that is required for the handling in VIAF.php) I haven't found if there is a parameter to set the 'simple' output like the AutoSuggest endpoint so pretty sure this will not work as in the current master nor develop :/

collectiveaccess:
Hmm. It worked last time I checked, but that was some time ago. I'll take a closer look.

PR #508, opened 2020-07-09
NOJIRA Replace glob call with Zend\Stdlib\Glob::glob

kehh:
* Running on docker using an Alpine PHP image resulted in warnings and failed globbing.
* https://github.com/docker-library/php/issues/719
Screenshot from 2020-07-09 23-33-36

### DIFF
--- a/app/helpers/printHelpers.php
+++ b/app/helpers/printHelpers.php
@@ -30,7 +30,9 @@
  * ----------------------------------------------------------------------
  */
 
- /**
+use Zend\Stdlib\Glob;
+
+/**
    *
    */
 	require_once(__CA_LIB_DIR__."/Print/PDFRenderer.php");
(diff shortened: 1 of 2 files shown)


yelto:
Alpine sounds like an interesting target. The commit might be a little aggressive though, I am not sure what the plan is going forward with Zend/Laminas (maybe a full refresh?)  but at the moment at least, providence/app/lib/Zend/Stdlib is not complete (no Glob in particular), so the build is broken..

collectiveaccess:
@yelto Try doing a composer update on your install. This PR work fine for me, but of course you do need to have the libraries installed.

yelto:
Oh, sorry it's been a while, used to build providence bare metal, I did not realize that composer was the canonical way as there is a semi-populated Zend/Stdlib/ in the code base. Can vagrant and editorconfig be relied upon as well? Thanks!

PR #509, opened 2020-07-09
NOJIRA fix warnings resulting from sizeof being executed on null values.

kehh:
Screenshot from 2020-07-10 01-14-36


collectiveaccess:
Look good. Thanks!

PR #536, opened 2020-07-26
Update expected description for wikipedia test.

### LINKED ISSUE #530: Broken test for get/AttributeGetTest.php
I just realized that the response from Wikipedia for the query on "Alexander Burr" that is in `get/AttributeGetTest.php` has changed (see and thus, the tests are now broken.

https://github.com/collectiveaccess/providence/blob/c38ca3c7e66a544f4c31ad6534212e06e942b789/tests/testsWithData/get/AttributeGetTest.php#L213

```

Failed asserting that '<p class="mw-empty-elt">\n
...
Burr fatally shot his political rival Hamilton in an 1804 duel, 
...
<li>Aaron Burr at Find a Grave</li></ul>' contains "Burr shot his political rival Alexander Hamilton in an 1804 duel".
 /Users/orestes/devel/providence/tests/testsWithData/get/AttributeGetTest.php:213
 /Users/orestes/devel/providence/vendor/phpunit/phpunit/phpunit:61

```

It is trivial to solve updating the expected text in the test, but I think the purpose of this test and some others is not testing that Wikipedia response is right, but that CollectiveAccess code is right. What about considering refactoring code a bit to mock the service calls to Wikipedia, so the test is about CA code and thus eliminating the external dependency on the tests?

I think there might be reason for not doing it, like making sure that those endpoints are operating and that the code is compatible with them, but I think we can refactor a short amount of the tests to interface with Wikipedia and some other tests to just mock the service.

What do you think?

miceno:
Fixes #530

### DIFF
--- a/tests/testsWithData/get/AttributeGetTest.php
+++ b/tests/testsWithData/get/AttributeGetTest.php
@@ -210,13 +210,13 @@ class AttributeGetTest extends BaseTestWithData {
 		$this->assertEquals('Coney Island', $this->opt_object->get('ca_objects.tgn'));
 		$this->assertStringContainsString('Aaron Burr', $this->opt_object->get('ca_objects.wikipedia'));
 		// subfield notation for "extra info"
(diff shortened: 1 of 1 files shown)


miceno:
closes #530

PR #548, opened 2020-08-07
Localize batch report status notice

yelto:


themes/default/views/batch/metadataimport/batch_results_html.php
```diff
@@ -105,14 +105,14 @@ function caUpdateBatchMetadataImportResultsReport($po_request, $pa_general, $pa_
 			foreach($pa_notices as $vn_id => $va_notice) {
 				switch($va_notice['status']) {
 					case 'SUCCESS':
-						$vs_buf .= "<li><em>".caEditorLink($po_request, $va_notice['label'], '', $pa_general['table'], $vn_id)."</em> (".$va_notice['idno']."): ".$va_notice['status']."</li>";
+						$vs_buf .= "<li><em>".caEditorLink($po_request, $va_notice['label'], '', $pa_general['table'], $vn_id)."</em> (".$va_notice['idno']."): "._t($va_notice['status'])."</li>";
```
collectiveaccess, line 108: These messages need to be localized upstream where the actual text is set. Localizing them like this means poedit won't pick up the string for translation.

yelto: The pass to pick up the strings to be translated will not pick up the values ('SUCCESS', etc) as strings to be translated, but if included otherwise in the po, they will be translated at runtime. Presenting the user with a raw item in an enumeration might not be the way to go in the first place, and we are exposing/compounding the issue by translating it ...economically. Ok, should just expand the switch to include explicitly the corresponding UI message for each value of $va_notice['status']. Thanks

collectiveaccess: Yes, that's true. But it's generally better practice to translate the strings where they're set.

yelto: Agreed, plus looking at the code the cleaner fix is cheap here.

PR #549, opened 2020-08-08
NOJIRA add Filesize to schema.

kehh:


collectiveaccess:
Thanks! This was on my todo list.

PR #551, opened 2020-08-10
Make REST API session lifetime configurable

artturimatias:
This make REST API session lifetime configurable in services.conf instead of hard coded 2 hours. Default is still 2 hours.


collectiveaccess:
Looks good. Thanks!

PR #554, opened 2020-08-12
Localize metadata import result status

yelto:
Per our previous conversation, with explicit strings for translation. Curiously, either the code was confused or I am, as the various 'status' values listed originally seem to pertain to batch media imports and not to metadata imports. If this is ok I can fix themes/default/views/batch/mediaimport/batch_results_html.php  which already contains a similar non-extractable _t() string.


collectiveaccess:
Yes, I think it's confused. Can you go ahead and clean up the other one as well? Thanks.

PR #561, opened 2020-08-18
NOJIRA Add an index for handler on ca_task_queue.

kehh:
* Allows for bulk updating task queue entries for a given handler.


collectiveaccess:
I'm going to have to merge this by hand with some other pending migrations. I will close this once I've done that, with a reference to the commit that implements this.

SethKaufman:
Added in separate pull request with expanded migration.

PR #572, opened 2020-09-02
Bug/ui autocomplete

the666bbq:
hi,

we found that the autocomplete lookup field for related fields (related collections, related entities,..) had side-effects: when the dropdown was shown, and the user scrolled outside this list, the body scrolled and the list got detached from the input field.
It appeared that the jquery autocomplete had no specific element referenced as the appendTo, thus defaults to appending it to the body.
I specified the appendTo as the 'parent' container via options.container option, and that keeps the dropdown in situ.


collectiveaccess:
This PR seems to have a bunch of other commits and is against master. I believe this issue is actually fixed in develop. I'll verify that this week.

the666bbq:
Thanks.
I thought I branched indeed from a clean master. Do you rather have PR's from another branch (develop?)?

(it's a one-line fix this one)

collectiveaccess:
Yes, it's better to base off of develop. What file is the change in? The current PR changes 782 files so it's unclear where the actual change is.

the666bbq:
assets/ca/ca.relationbundle.js

collectiveaccess:
Got it. I'm manually adding this change to develop. Thanks.

PR #582, opened 2020-09-11
Refactor to use AAT query mocks

miceno:
Refactor testing code to substitute remote API calls to AAT by mocks. It will reduce external dependency on AAT API service.

Nevertheless, I kept one of the end-to-end tests to make sure there is a test that properly tests AAT service is up and running.


kehh (commented):
Looks good to me

PR #583, opened 2020-09-14
Use ZendFramework from Composer

### LINKED ISSUE #568: Using ZendFramework1 from packagist
Hi,

Have you consider adding Zend Framework as a dependency on the vendor library?

This way you will get rid of all the `require`, `include`, etc.

As of now, Zend Framework 1.12 is deprecated, but there are two available packages in packagist: the original one (1.12.20) and the one forked for Magento (1.14) supports PHP 7.3 (https://github.com/magento/zf1/releases/tag/1.14.2)

What do you think?

miceno:
Update dependencies to remove obsolete ZendFramework dependencies (other deprecated packages remain).

Uses Magento ZendFramework 1.13.1 (yes, that exact version), since MZF 1.14 removed Lucene and other components.

Closes #568


miceno:
Giving up.

collectiveaccess:
Sorry, we've been distracted with things. Will review and get back to you by Monday.

collectiveaccess:
This looks good. Thank you for contributing this.

PR #601, opened 2020-09-29
NOJIRA Don't spam ElasticSearch checking for the record's existence for _every_ _single_ _field_.

kehh:


collectiveaccess:
Good idea :-/

kehh:
We were wondering why ES was groaning when editing lots of records.

PR #621, opened 2020-10-07
Add abilityto disable  mediaDirRecursiveExpensiveCheck

vatoz:
mediaDirRecursiveExpensiveCheck can be costly on large installations (25000 folders in media), Configuration check then fails on silent timeout.
I have added config to bypass this test (with bypass disabled on default).


collectiveaccess:
Good idea. Thanks!

PR #651, opened 2020-10-30
NOJIRA Fix metadata element usage counts to reflect changes in bundle…

kehh:
… name.

* Matches either ca_attribute_ or ca_table_name.


collectiveaccess:
Thanks for this one. It has been on my todo list forever.

PR #680, opened 2020-11-19
Generate gettext Portable Object Template (POT) and apply to po files

vatoz:
PO files at repository are somehow outdated as well as Transifex translation sources.

Here i offer small script able to generate POT  file (usable with many tools involved in translation)  and generated file.

Script include  strings from php and conf files (eg. navigation), genereate pot file  and apply it to all po files.
Afterwards it will generate mo files. 

Changed po files are not part of this PR.

Known bug: It doesnt cope with app.conf ( curently in script its removed by  grep).
Possible enhancement: It might be wise to move this script to support folder?


collectiveaccess:
Thanks for this! I agree we should move this to the support directory. Do you want to resubmit with that move? Or shall I modify it?

vatoz:
I can prepare it.
I believe, that this tool is necessary to have translation source in
Transifex in sync with code.
Dne so 28. 11. 2020 4:53 uživatel CollectiveAccess <notifications@github.com>
napsal:

collectiveaccess:
That's excellent. Thank you. This tool will be very handy. Sync'ing Transifex by hand has meant it is always quite out of date.

vatoz:
I will prepare it as separate pull request.

PR #700, opened 2020-11-28
PR-653 Fix broken Font-Awesome update

SethKaufman:
PR adds missing font-awesome files, adjusts icon classes and moves Mirador CSS load to avoid conflict with Font-Awesome 5 update.


kehh:
Oops sorry

PR #701, opened 2020-11-29
Solve a typo on SearchIndexer

miceno:
While solving undefined reference, a typo was introduced.

This PR solves it.


collectiveaccess:
Ugh. Thanks for catching these!

PR #713, opened 2020-12-02
NOJIRA Add a way of filtering AAT information service values.

kehh:


kehh:
@collectiveaccess if you're happy with this approach let me know and I'll fill it out for the other Getty attribute types.

miceno:
Great! A PR with tests! Thank you for improving the quality of the code!

SethKaufman:
Sorry to let this lie for so long. It looks good. Let me know if you still want to expand it.

kehh:
Yes - I've got another tweak to push that allows altering the query to find children of more than one term.

I also have a separate getty enhancement that:
1.  adds cacheing of the responses and
2. appends a wildcard to the search query to make it behave more like a standard autocomplete.
 
Do you want me to put those into separate PR's or alongside this one? I'm going to go with separate although it does touch very similar code.

collectiveaccess:
I would submit the Getty mods separately if it's not too much hassle.

PR #735, opened 2020-12-15
handle undefined __CA_SYSTEM_GUID__

tacman:


collectiveaccess:
Thanks for catching this. It got pulled over from 1.8. The better solution would be to define it properly... which I shall do.

PR #751, opened 2020-12-21
Test idno_regexes

vatoz:
Something is wrong in configuration parsing of conf files, 
i have found it on idno_regexes at search.conf.  

Code  from SearchEngine.php  expects something like 
array(  
 'ca_objects' = array(	'[\d]{4}\.[\d]{1,5}\.[\d]{0,5}') 
)
but it gets something like 
array (
  0 => 'ca_objects = [[d]{4}\\.[d]{1,5}\\.[d]{0,5}',
)

This PR adds just error reporting, it doesn't include fix.

### DIFF
--- a/tests/lib/ConfigurationTest.php
+++ b/tests/lib/ConfigurationTest.php
@@ -150,7 +150,7 @@ class ConfigurationTest extends PHPUnit_Framework_TestCase {
 			$va_regexes = $o_config->get("idno_regexes");
 			$this->assertTrue(is_array($va_regexes));
 			$this->assertArrayHasKey("ca_objects",$va_regexes);
-			$this->assertEquals($va_regexes["ca_objects"][0], "[\d]{4}\.[\d]{1,5}\.[\d]{0,5}");
+			$this->assertEquals('[\d]{4}\.[\d]{1,5}\.[\d]{0,5}',$va_regexes["ca_objects"][0]);
 	}
 
 
--- a/tests/lib/ConfigurationTest.php
+++ b/tests/lib/ConfigurationTest.php
@@ -37,13 +37,13 @@ class ConfigurationTest extends PHPUnit_Framework_TestCase {
 	public function testScalars() {
 		$o_config = new Configuration(__CA_BASE_DIR__.'/tests/lib/data/test.conf', false, true);
 
-		$this->assertEquals($o_config->get('a_scalar'), 'Hi there');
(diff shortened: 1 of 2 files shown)


collectiveaccess:
Double backslashes ("\\") are required here. Single "\" characters would be parsed as escaping what comes next. So \. = . while \\. = \.

We need those backslashes to be passed as literals in the regex so PHP can use them. I'll check the config files and make sure double backslashes are used.

Thanks for catching this.

vatoz:
Well, lines included in my commit 
idno_regexes = [
	ca_objects = [
		"[\d]{4}\\.[\d]{1,5}\\.[\d]{0,5}"
	]
]

was taken from providence/master app/conf/search.conf

collectiveaccess:
I've added backslashes in search.conf. The entries in app.conf are ok.

vatoz:
@collectiveaccess I have updated test.conf  with double backslashes. 

But the test still fail.

PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.....F                                                              6 / 6 (100%)

Time: 289 ms, Memory: 34.50MB

There was 1 failure:

1) ConfigurationTest::testGreps
Failed asserting that an array has the key 'ca_objects'.

/home/apps/collectiveaccess/phpunit_testing/tests/lib/ConfigurationTest.php:152

FAILURES!
Tests: 6, Assertions: 61, Failures: 1.

vatoz:
Hello @collectiveaccess ,
i have found, that part of problem with config  wass in different handling of lists and associative lists.
Its  changed in test conf.
So, my test now fails just on doble backslashes processing.

BTW. I have also reorderd parameters to assertEquals. According to https://phpunit.readthedocs.io/en/9.3/assertions.html#assertequals first is expected value and second is actual value. Now it's last part of PR.


PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.....F                                                              6 / 6 (100%)

Time: 297 ms, Memory: 34.50MB

There was 1 failure:

1) ConfigurationTest::testGreps
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'[\d]{4}\.[\d]{1,5}\.[\d]{0,5}'
+'[d]{4}.[d]{1,5}.[d]{0,5}'

/home/apps/collectiveaccess/phpunit_testing/tests/lib/ConfigurationTest.php:153

FAILURES!
Tests: 6, Assertions: 62, Failures: 1.

collectiveaccess:
I see the problem. A fix will be pushed shortly.

PR #791, opened 2021-01-17
NOJIRA fix purifier call when logging an error message.

kehh:


app/lib/Logging/Eventlog.php
```diff
@@ -77,13 +77,13 @@ public function log($pa_entry) {
...
 			}
-			
+			$purifier = new HTMLPurifier();
 			$this->o_db->query("
 				INSERT INTO ca_eventlog 
 				(date_time, code, message, source)
 				VALUES
 				(unix_timestamp(), ?, ?, ?)
-			", $pa_entry["CODE"], $this->purify($pa_entry["MESSAGE"]), $pa_entry["SOURCE"]);
+			", $pa_entry["CODE"], $purifier->purify($pa_entry["MESSAGE"]), $pa_entry["SOURCE"]);
```
kehh, line 86: Although I think this might 'double purify' the message, once on the way in to the event log and then once on the way out again further down in the file. However this _does_ fix a fatal error:
```
Fatal error: Uncaught Error: Call to undefined method Eventlog::purify() in /var/www/html/tas-providence/app/lib/Logging/Eventlog.php:86 Stack trace: #0 /var/www/html/tas-providence/app/lib/Controller/Request/RequestHTTP.php(895): Eventlog->log(Array) #1 /var/www/html/tas-providence/app/controllers/system/AuthController.php(66): RequestHTTP->doAuthentication(Array) #2 /var/www/html/tas-providence/app/lib/Controller/RequestDispatcher.php(281): AuthController->DoLogin(NULL) #3 /var/www/html/tas-providence/app/lib/Controller/AppController.php(113): RequestDispatcher->dispatch(Array) #4 /var/www/html/tas-providence/index.php(121): AppController->dispatch(true) #5 {main} thrown in /var/www/html/tas-providence/app/lib/Logging/Eventlog.php on line 86
```

collectiveaccess:
Incoming purification can be turned off (and often is for performance reason). Purification was added here to ensure it's always filtered no matter what. But perhaps a better approach would be to force purification on eventlog entries no matter what's configured.

PR #796, opened 2021-01-20
Quick fix for Set list by user headings

yelto:
Set list by user is a nested iteration, as it iterates over users, and for each user over the user's sets. The code sort of mixes both iterations, the pagination seems to happen only on users, instead of on sets or on users and sets. The shoehorning of both iterations into presentation  code meant for one results in the wrong heading being shown. This hack does not address the pagination issue, it just postpones setting the 'user' label into type_name_* for pagination purposes until after these variables have produced the right heading for the real set type. At least users are slightly less confused...

### DIFF
--- a/app/controllers/manage/SetController.php
+++ b/app/controllers/manage/SetController.php
@@ -240,8 +240,6 @@
             
          
             $this->view->setVar('set_list', $va_set_list);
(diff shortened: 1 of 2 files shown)


collectiveaccess:
Thanks for this quick fix. We'll have to take a closer list at what is going on. For now we'll merge this in.

PR #799, opened 2021-01-24
use [] instead of {} for getting character

tacman:
using brackets doesn't work for PHP 8, throws deprecation warning in PHP 7.4


collectiveaccess:
Please submit this against develop, not master. Thanks.

PR #811, opened 2021-01-30
Localize required_field_marker

yelto:
Localize the default text in required_field_marker, for instances where show_required_field_marker is set to 1.

The text to be translated would not be picked up if maintained verbatim  (that is, _t((Required)) would not be localized at runtime for some reason tied to the use of parenthesis). Not knowing the escape rules (if any), I've cowardly moved the string to square brackets, with _t([Required]) working as expected. Installs using the default string will notice a change from (Required) to [Required]. Hope it helps anyway.


collectiveaccess:
I think square brackets work better actually.

PR #814, opened 2021-02-01
Fix class name for Llk

tacman:
Perhaps Hoa has been refactored.


collectiveaccess:
What is the problem you're trying to solve here. We have not seen any errors with Expression parsing.

tacman:
Static code analysis throws a warning, because the class hint isn't valid.

PR #824, opened 2021-02-11
make $opn_value_id protected

tacman:
Since MediaAttributeValue accesses this, it can't be private.


tacman:
Agreed.  We also need a SetValueId() method.

PR #825, opened 2021-02-11
array_shift requires a variable passed by reference

tacman:
replace it with [0], since we just want the first element, not shifting the array.


collectiveaccess:
Please don't submit against master. And what is the problem you're trying to solve here?

tacman:
Drat!  I thought I was in develop.

I'm getting rid of php errors -- array_shift needs an array passed by
reference.  Otherwise, it shows up as an error and analysis tools fail.

PR #861, opened 2021-03-12
Add support for `Custom` paper size when printing.

kehh:
* Enable wkhtmltopdf support for custom page size and print media queries.
  * use --page-height and --page-width for custom page size
  * Use --page-size for standard page size
* Fix calculation of points when initial measurement is in mm (25.4 mm in an inch).
* More precision when calculating metric lengths from point.

This is specified as follows:
```
 * @pageSize Custom(89mm,28.5mm)
```


kehh:
[barcode.pdf](https://github.com/collectiveaccess/providence/files/6127376/barcode.pdf)

PR #872, opened 2021-03-22
NOJIRA Allow representations to be set during editItem handler in the item service

kehh:
* Also allow representations to be saved when there are no relationships.


app/lib/Service/ItemService.php
```diff
@@ -903,29 +903,11 @@ public function addItem($ps_table=null, $pa_data=null) {
 					}
 				}
 			}
+		}
```
kehh, line 906: This was previously inside the 'related' processing, and representations don't require related records so I've moved it just outside that `if`

PR #883, opened 2021-03-26
PROV-3101 Trigger filters on child attributes inside a container.

kehh:
This PR further fixes an issue where modification metadata alerts don't get triggered for attributes inside a container.


collectiveaccess:
Argh. I didn't think about trigger values. Thanks for this.

PR #897, opened 2021-04-06
PROV-3108 Fix tac regressions (barcodes + importer)

SethKaufman:
PR resolves recently introduced errors due to elimination of old bar code libraries and no preload for data importer refinery manager.


kehh:
@collectiveaccess this didn't have a composer change in the MR. Was that meant to be in there?

collectiveaccess:
There is a composer.json change in there. You don't see it?

PR #924, opened 2021-04-27
PROV-3122 Don't truncte ElasticSearch index when installer page in loaded unless install is actually going to happen

SethKaufman:
PR addresses issue where preinstall tasks are run any time the installer page is loaded. The is unnecessary but benign, unless ElasticSearch indexing is used, in which case the preinstall task handler will truncate the ES index before evaluating whether the install is actually allowed.

This PR makes preinstall tasks contingent on evaluation of installation validity and adds checks in a couple of other places for good measure.

### DIFF
--- a/install/inc/Installer.php
+++ b/install/inc/Installer.php
@@ -405,7 +405,7 @@ class Installer {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


kehh (approved):
Yup this makes sense.

PR #954, opened 2021-05-25
NOJIRA Fix set batch editing when opening set from search results.

kehh:
* Previously the set id parameter passed didn't match the url pattern
used in 1b30f3d5ba37489f95c8331d294ef43a6739aae9
* Fixes an issue where if you click onto a different screen other than
the default screen in the batch edit UI you get an
ApplicationException::('Invalid target')


collectiveaccess:
Thanks for catching that one. I had thought I'd clicked on everything at least once. Obviously not :-/

PR #972, opened 2021-06-04
add ADBCTable entry awareness to tokenizing

the666bbq:
Hi, when an author wants to enter a date like "1000 voor Christus" the date is invalid regardless the entry in the ADBCTable of the language TimeExpressionParser configuration. I tried quoting too, like de_DE.lang but that didn't validate either.
`ADBCTable = { ..., voor Christus = bc, ..}`

It seems the date is tokenized to [1000,voor,Christus] and those individuals are then missed by the era_lookup.

I added parsing of the ADBCTable entries first, adding those to the normal tokenizing of the leftover ps_expression resulting to [1000, voor Christus], with as a result a valid date.

### DIFF
--- a/app/lib/Parsers/TimeExpressionParser.php
+++ b/app/lib/Parsers/TimeExpressionParser.php
@@ -1671,8 +1671,28 @@ class TimeExpressionParser {
 	# Lexical analysis
 	# -------------------------------------------------------------------
 	private function tokenize($ps_expression) {
+
+		$arr_tokens = $this->multiwordADBCtokens($ps_expression);
+		$ps_expression = $arr_tokens['expression'];
+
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Why not just rewrite the expression in preprocess()? It seems to me that that would be simpler and easier for others patching the parser later to debug.

the666bbq:
> Why not just rewrite the expression in preprocess()? It seems to me that that would be simpler and easier for others patching the parser later to debug.

thx for that info, didn't notice the preprocess hook for TimeExpressionParser. So, I guess that would be a hookTimeExpressionParserPreprocessAfter and than replace/return "1000 voor Christus" to "1000 BC". The value entered will still be the original "1000 voor Christus" then ?

the666bbq:
ok, thx, preprocesshook work-around works

PR #985, opened 2021-06-09
Feature/configure sets per page

the666bbq:
opn_items_per_page was still a static value in the SetController. Changed to a configurable setting, where I tried to mimic the other items_per_page_default_for_[concept]_search in app.conf that existed already for the other concepts but not ca_sets.


collectiveaccess:
Looks good. Thanks.

PR #1006, opened 2021-06-28
Allow adding exporter worksheets without dnd

raboof:
I haven't checked it still works with dnd.

After uploading, the table is not refreshed, but I'm not sure whether
that worked before either ;).

Probably needs some styling as well.

Kept it as draft for now because of the above issues, but feedback welcome!

### DIFF
--- a/themes/default/views/manage/export/exporter_list_html.php
+++ b/themes/default/views/manage/export/exporter_list_html.php
@@ -56,6 +56,7 @@ if (!$this->request->isAjax()) {
 	
 	<div id="exporterUploadArea" >
 		<span class="exporterUploadText"><?php print _t("Drag exporter worksheets here to add or update"); ?></span>
+		<input type="file" id="exporterUpload">
 	</div>
 <?php
 }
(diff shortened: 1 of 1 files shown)


collectiveaccess:
What problem is being solved here? I don't recall the motivation... sorry.

raboof:
> I don't recall the motivation... sorry.

No worries, we haven't discussed this before ;)

I noticed the only way to upload a new exporter worksheet is currently via drag-and-drop. That is not always convenient: I didn't happen to have a window to 'drag from' open, and more generally users with particular accessibility preferences/needs might not have drag-and-drop easily available.

So I figured it'd be nice to give the option of uploading files with the standard file selection dialog as well, aside from drag-and-drop.

What do you think?

collectiveaccess:
That's a really good idea. It has always bothered me that drag-and-drop was the only way. We'll need to do it for importer mappings as well though :-/

collectiveaccess:
PS. I'll take a look at this, fix up styling and merge.

collectiveaccess:
This is still on the todo list...

raboof:
thanks for the update, no rush ;)

PR #1007, opened 2021-06-28
Check the media upload tmp dir is writable

### LINKED ISSUE #965: check upload_tmp_dir is writable on startup
On startup, providence already checks if a number of directories are present and writable. This is very helpful. It looks like the php `upload_tmp_dir` is not checked yet, but is needed for uploading media, so that would be nice to add.

(encountered on 1.7.11, haven't checked master yet)

raboof:
Fixes #965


collectiveaccess:
Looks good. Thank you!

PR #1016, opened 2021-07-01
Trigger metadata alert filters for multi value containers

kehh:
* Previously _if_:
  * you have an element inside a container and
   * the container is multi valued
* _then_ filtered alerts do not get triggered.
* This change compares the individual values for an attribute rather
than concatenated values.


collectiveaccess:
Looks good. Thank you for working on this!

PR #1019, opened 2021-07-02
Remove die statement

kehh:
I just did a pull from `develop` and this didn't look safe.


collectiveaccess:
Yes thanks. That got sucked in by mistake earlier this evening.

PR #1035, opened 2021-07-07
Fix syntax error for the places search builder

kehh:
* Copied from ca_objects_search_builder_html.php


collectiveaccess:
Oops. That'll teach me to work at night. Sorry about that.

PR #1064, opened 2021-07-26
Replace colours with slightly more accessible colours

kehh:
We've been asked to tweak the stylesheets for slightly more accessible colours just to add a bit more contrast.

* red and co: => https://coolors.co/eb0000
* link blue => #138290

### DIFF
--- a/themes/default/css/base.css
+++ b/themes/default/css/base.css
@@ -38,7 +38,7 @@ textarea {
     
 }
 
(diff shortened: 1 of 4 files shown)


kehh:
Happy to do this in our own theme, but figured you might want this

collectiveaccess:
Yeah, please. Let's give it a try.

PR #1094, opened 2021-08-25
Tweaks to data exporter

kehh:
* Support exporting of contents of relationship tables
* Tweak to column headers for CSV
  * Properly default the column name if it doesn't have a display label
  * Support CSV_delimiter and CSV_enclosure for column headings


collectiveaccess:
Looks good. Thanks!

PR #1112, opened 2021-09-07
Don't clobber the server when there are too many jobs pending/processed

kehh:
* Adds a configurable limit to trackProcessingWidget
* Formats dates using localised dates


collectiveaccess:
Thanks. I've been meaning to do this for ages.

PR #1118, opened 2021-09-10
Only save record on prepopulate if there have been any changes.

kehh:
We're currently using prepopulate with an `edit` configuration which is resulting in change log entries for records every time you view them rather than when you actually edit them. 

I think this change will only save the record if there have been any actual changes as a result of the prepopulate.

### DIFF
--- a/app/plugins/prepopulate/prepopulatePlugin.php
+++ b/app/plugins/prepopulate/prepopulatePlugin.php
@@ -525,15 +525,16 @@ class prepopulatePlugin extends BaseApplicationPlugin {
 		}
 
 
-		if(isset($_REQUEST['form_timestamp']) && ($_REQUEST['form_timestamp'] > 0)) { $_REQUEST['form_timestamp'] = time(); }
-		$t_instance->update(['force' => true, 'hooks' => false]);
-
-		if($t_instance->numErrors() > 0) {
-			foreach($t_instance->getErrors() as $vs_error) {
-				Debug::msg("[prepopulateFields()] there was an error while updating the record: ".$vs_error);
+		if ($t_instance->getChangedFieldValuesArray()) {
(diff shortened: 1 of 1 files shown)


collectiveaccess:
I've noticed this issue as well. I'll look it over and test it soon.

app/plugins/prepopulate/prepopulatePlugin.php
```diff
@@ -525,15 +525,16 @@ public function prepopulateFields(&$t_instance, $pa_options=null) {
 		}
 
 
-		if(isset($_REQUEST['form_timestamp']) && ($_REQUEST['form_timestamp'] > 0)) { $_REQUEST['form_timestamp'] = time(); }
-		$t_instance->update(['force' => true, 'hooks' => false]);
-
-		if($t_instance->numErrors() > 0) {
-			foreach($t_instance->getErrors() as $vs_error) {
-				Debug::msg("[prepopulateFields()] there was an error while updating the record: ".$vs_error);
+		if ($t_instance->getChangedFieldValuesArray()) {
```
collectiveaccess, line 528: Won't this always evaluate to true? getChangedFieldValuesArray returns an array, which will be empty (but array-is) if nothing has changed. Need to check size of array here?

kehh: empty arrays are falseish:
```
 php -a
Interactive shell

php > if ([]) { echo 'true';} else { echo 'false';}
false
```
But happy to put a count around it

collectiveaccess: My preference would be to test count explicitly

kehh: Sorted

collectiveaccess: Thank you!

PR #1126, opened 2021-09-14
Only set context for idno if it isn't already set.

kehh:
* Fixes warning about not being able to set the type_id if that has
already been set for a record.


collectiveaccess:
Was this causing a specific issue in the UI?

kehh:
With `__CA_ENABLE_DEBUG_OUTPUT__` this was giving a warning about not being able to set a type_id for a record that's already been set. So this would not normally be displayed to users without debug turned on. PR only set's the value if it hasn't been set.

PR #1131, opened 2021-09-17
Option to order AAT results by term

RedHaxor:
A simple addition to allow admins to select to order the results by the term or leave it in the default order.

### DIFF
--- a/app/lib/Plugins/InformationService/AAT.php
+++ b/app/lib/Plugins/InformationService/AAT.php
@@ -62,7 +62,18 @@ class WLPlugInformationServiceAAT extends BaseGettyLODServicePlugin implements I
 			'label' => _t('Additional sparql suffix'),
 			'description' => _t('Applied after the initial search. Useful to combine filters. For example to limit to children of particular terms enter "?ID gvp:broaderPreferredExtended ?Extended FILTER (?Extended IN (aat:300261086, aat:300264550))"')
 		];
-
+		$g_information_service_settings_AAT['orderBy'] = [
+			'formatType' => FT_TEXT,
+			'display_Type' => DT_FIELD,
+			'default' => '',
+			'width' => 90, 'height' => 1,
+			'label' => _t("Order by"),
+			'description' => _t("How the auto complete list will be ordered. Default is how AAT return the results. Label is by the visible label"),
+			'options' => array(
+				_t("Default") => "Order",
+				_t("Label") => "TermPrefLabel"
+			),
+		];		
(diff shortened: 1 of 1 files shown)


app/lib/Plugins/InformationService/AAT.php
```diff
@@ -62,7 +62,18 @@ public function __construct() {
 			'label' => _t('Additional sparql suffix'),
 			'description' => _t('Applied after the initial search. Useful to combine filters. For example to limit to children of particular terms enter "?ID gvp:broaderPreferredExtended ?Extended FILTER (?Extended IN (aat:300261086, aat:300264550))"')
 		];
-
+		$g_information_service_settings_AAT['orderBy'] = [
+			'formatType' => FT_TEXT,
+			'display_Type' => DT_FIELD,
```
kehh, line 67: I think this should be
```
displayType => DT_SELECT,
```

app/lib/Plugins/InformationService/AAT.php
```diff
@@ -62,7 +62,18 @@ public function __construct() {
...
+			'default' => '',
+			'width' => 90, 'height' => 1,
+			'label' => _t("Order by"),
+			'description' => _t("How the auto complete list will be ordered. Default is how AAT return the results. Label is by the visible label"),
+			'options' => array(
+				_t("Default") => "Order",
```
kehh, line 73: Maybe make these options configurable in linked_data.conf?

Something like:
```
aat = {
	sortOptions = {
		_(Default) = Order,
		_(Label) = TermPrefLabel,
	},
	...
}
```

But that's not required for this to work just optional.

collectiveaccess:
That's a nice idea, if someone wants to make the change and submit another PR. But I kind of doubt there'll be many other options actually used.

app/lib/Plugins/InformationService/AAT.php
kehh: (You'll only be able to read the config after the `parent::__construct()` and it'll be available as:
```
		$va_config = $this->opo_linked_data_conf->get($this->getConfigName());
		$g_information_service_settings_AAT['orderBy']['options'] = $va_config['sortOptions'] ?? [];
```

kehh: Oops I just saw this was merged ;)

RedHaxor:
Only other options I could think of was to order in DESC, but I figured people who want it sorted will want it alphabetically in Ascending order.

collectiveaccess:
I wouldn't over-engineer this. Just my 2 cents :-/

PR #1136, opened 2021-09-20
Tweaked bundle code display

RedHaxor:
Bundle codes were showing incorrectly for fields outside of containers and for containers within containers.
This fixes both of these issues by not showing the bundle code on non-container fields (the bundle header will cover these) and by gathering any containers within containers during the loop and adding them to $container_bundle_code.  
I have only tested the container_bundle_code on 2 deep containers, will need testing with more nested containers.

### DIFF
--- a/app/lib/BaseModelWithAttributes.php
+++ b/app/lib/BaseModelWithAttributes.php
@@ -1819,15 +1819,16 @@
 				}
 			}
 			
-			$container_bundle_code = '';
+			$t_element_datatype = $t_element->get('datatype');
+			$t_element_code = $t_element->get('element_code');
+			$table_name = $this->tableName();
+			$show_bundle_codes = $po_request->user->getPreference('show_bundle_codes_in_editor');
+			
 			foreach($va_element_set as $va_element) {
 				$va_element_info[$va_element['element_id']] = $va_element;
-				if (($va_element['datatype'] == 0) && ($va_element['parent_id'] > 0)) { 
-					// Gather bundle codes for containers within containers.
-					$container_bundle_code = "{$container_bundle_code}.".$va_element['element_code'];
-					continue; }
-
 				
+				if (($va_element['datatype'] == 0) && ($va_element['parent_id'] > 0)) { continue; }
+	
 				$va_label = $this->getAttributeLabelAndDescription($va_element['element_id']);
 
 				if(!isset($va_elements_without_break_by_container[$va_element['parent_id']])){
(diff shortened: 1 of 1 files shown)


app/lib/BaseModelWithAttributes.php
```diff
@@ -1849,10 +1853,12 @@ public function getAttributeHTMLFormBundle($po_request, $ps_form_name, $pm_eleme
 				
 				$label = (sizeof($va_element_set) > 1) ? $va_label['name'] : '';
 				
-				$show_bundle_codes = $po_request->user->getPreference('show_bundle_codes_in_editor');
-				$bundle_code = $this->tableName().'.'.$t_element->get('element_code').'.'.$va_element['element_code'];
+				if($t_element->get('datatype') == 0){ //Only show field level bundle codes inside containers
```
collectiveaccess, line 1856: It would be good to assign datatype to a variable outside of the loop and use that in the comparison

app/lib/BaseModelWithAttributes.php
```diff
@@ -1849,10 +1853,12 @@ public function getAttributeHTMLFormBundle($po_request, $ps_form_name, $pm_eleme
 				
 				$label = (sizeof($va_element_set) > 1) ? $va_label['name'] : '';
 				
-				$show_bundle_codes = $po_request->user->getPreference('show_bundle_codes_in_editor');
-				$bundle_code = $this->tableName().'.'.$t_element->get('element_code').'.'.$va_element['element_code'];
+				if($t_element->get('datatype') == 0){ //Only show field level bundle codes inside containers  
+					$show_bundle_codes = $po_request->user->getPreference('show_bundle_codes_in_editor');
```
collectiveaccess, line 1857: It would be a good idea to move this outside of the loop

app/lib/BaseModelWithAttributes.php
```diff
@@ -1819,9 +1819,13 @@ public function getAttributeHTMLFormBundle($po_request, $ps_form_name, $pm_eleme
 				}
 			}
 			
+			$container_bundle_code = '';
 			foreach($va_element_set as $va_element) {
 				$va_element_info[$va_element['element_id']] = $va_element;
-				if (($va_element['datatype'] == 0) && ($va_element['parent_id'] > 0)) { continue; }
+				if (($va_element['datatype'] == 0) && ($va_element['parent_id'] > 0)) {
```
collectiveaccess, line 1825: Nested containers are used for visual grouping but aren't actually part of the bundle code. It's always two level: container => sub-element. So we don't want to be appending those sub-container codes here, just the code from the top-level container.

collectiveaccess (changes requested):
Thanks for this! It's almost there.

collectiveaccess:
Looks good to me.

collectiveaccess (approved):
Looks good to me.

PR #1312, opened 2021-12-28
caExtractTagsFromTemplate : broken for marc

gautiermichelin:
Hi CA Team,

Adding the "/" to the not-to-be-removed parts of a template, as "^701/a" is a valid template, handled by method get inside MarcDataReader 
```
	public function get($ps_spec, $pa_options=null) {
		// [...]
		list($ps_code, $ps_subcode, $ps_indicators) = explode('/', $ps_spec);
```

I know these times marc tends to be more and more replaced by XMLmarc, but there are still some uses for example with Z9350 when SUTRS is not available.

Gautier


gautiermichelin:
Bump for this little one. Is there somewhere the / would be problematic inside templates ? 
I can change eventually the marc coding to use ^701a as a template for import, what do you think ?

collectiveaccess:
It seems ok.

Issue #1324, opened 2022-03-02
Thumbnail generator


vatoz:
Since  commit  f1934194ed35fc99dd132b1ef649789c7beddb76   (in master) generate thumbnails ceased to work.

I have tracked two different errors.
First one is this - it was calling GD function on machine with GD disabled.
I dont have a fix for it.

`[01-Mar-2022 15:51:03 Europe/Prague] PHP Fatal error:  Uncaught Error: Call to undefined function imagecreatefromjpeg() in /var/www/providence/app/lib/Parsers/TilepicParser.php:1575
Stack trace:
#0 /var/www/providence/app/lib/Parsers/TilepicParser.php(326): TilepicParser->encode_gd()
#1 /var/www/providence/app/lib/Plugins/Media/ImageMagick.php(834): TilepicParser->encode()
#2 /var/www/providence/app/lib/Media.php(284): WLPlugMediaImageMagick->write()
#3 /var/www/providence/app/lib/Plugins/TaskQueueHandlers/mediaproc.php(360): Media->write()
#4 /var/www/providence/app/lib/TaskQueue.php(328): WLPlugTaskQueueHandlermediaproc->process()
#5 /var/www/providence/app/lib/Utils/CLIUtils.php(738): TaskQueue->processQueue()
#6 /var/www/providence/support/bin/caUtils(167): CLIUtils::process_task_queue()
#7 {main}
  thrown in /var/www/providence/app/lib/Parsers/TilepicParser.php on line 1575
`

Second  and third error (found after changing installed packages on debug server)  -  Instead  of exec, new code uitilizes new systemHelper, which was not included in your commit.

`[02-Mar-2022 15:10:03 Europe/Prague] PHP Fatal error:  Uncaught Error: Call to undefined function caExec() in /var/www/providence/app/lib/Parsers/TilepicParser.php:366
Stack trace:
#0 /var/www/providence/app/lib/Parsers/TilepicParser.php(801): TilepicParser->_graphicsMagickRead()
#1 /var/www/providence/app/lib/Parsers/TilepicParser.php(332): TilepicParser->encode_graphicsmagick()
#2 /var/www/providence/app/lib/Plugins/Media/GraphicsMagick.php(830): TilepicParser->encode()
#3 /var/www/providence/app/lib/Media.php(284): WLPlugMediaGraphicsMagick->write()
#4 /var/www/providence/app/lib/Plugins/TaskQueueHandlers/mediaproc.php(360): Media->write()
[...]

PR #1325, opened 2022-03-02
Repair thumb generator - caExec calls

vatoz:
Works for me, but it is probably wrong/may break other parts. For description see issue #1324


vatoz:
Repaired by different commit.

PR #1331, opened 2022-03-31
allow empty separator in ID numbering

peterjanssens:
Hi,

Following the docs (https://manual.collectiveaccess.org/configuration/mainConfiguration/multipart_id_numbering.html) the separtor is mandatory, but should be allow to be empty.
Most (if not all) Numbering will have a fallback value for unset values so I assume this BaseModel forcing the dash/minus for emtpy values is no longer valid logic ?

I flaws our (valid) Multipart ID definition of constant (B) with serial (000001) with empty separator from wanted B000001 to unwanted B-00001

### DIFF
--- a/app/lib/BundlableLabelableBaseModelWithAttributes.php
+++ b/app/lib/BundlableLabelableBaseModelWithAttributes.php
@@ -475,7 +475,7 @@ class BundlableLabelableBaseModelWithAttributes extends LabelableBaseModelWithAt
 		if ($vs_idno_fld && (!isset($va_duplicate_element_settings[$vs_idno_fld]) || (isset($va_duplicate_element_settings[$vs_idno_fld]) && ($va_duplicate_element_settings[$vs_idno_fld] == 1)))) {
 			$vb_needs_suffix_generated = false;
 			if (method_exists($this, "getIDNoPlugInInstance") && ($o_numbering_plugin = $this->getIDNoPlugInInstance())) {
-				if (!($vs_sep = $o_numbering_plugin->getSeparator())) { $vs_sep = '-'; }	// Must have a separator or you can get inexplicable numbers as numeric increments are appended string-style
+				$vs_sep = $o_numbering_plugin->getSeparator();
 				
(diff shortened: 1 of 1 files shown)


collectiveaccess:
You can have a multipart number without a separator, so long as it only has a single element. With multiple elements, the lack of a separator prevents consistent segmentation of the value and unpredictable results in many numbering configurations. 

This was added to handle those cases, as if the identifier of the duplication is invalid duplication will fail.  Probably the better way to deal with it would be to try it without a forced separator and fall back to a default separator if the initial try fails.

peterjanssens:
But a separator makes little sense when you have one element, no :)  ?

``` {
        ...
 separator =,
			
			elements = {
				accession_type = {
					type = CONSTANT, 
					value = B, 
					default = B, 
					width = 1, 
					description = _(Accession type), 
					editable = 0
				},
				object_number = {
					type = SERIAL,
					width = 8,
					description = _(Object identificatiecode),
					editable = 0,
					table = ca_objects,
					field = idno,
					sort_field = idno_sort,
					zeropad_to_length = 8
				}
			}
```

It does work fine for "new" creations where the next next B0000xx is calculated correctly. And the next identifier parts are calculated correctly for duplications too, it's just that the separator gets 'overruled' and results in B-0000xx.
The check for empty seems redundant at that point when the handlers had their fallback separator already.

Maybe I don't understand the difference between duplication and normal workflow for the identifier.

It's probably easier to explode the numbering on a separator, but don't all the parts have widths ?

I think the only other way around is via plugin then either via conditional replacing of the original dash (or another non-empty separator) on some types (because we too can have hybrid formats), or through unconditional replacement of a more exotic __SEPPPARATOR__.
If you know another way we'll be happy to try that.

But it seems to work fine without that current extra fallback to - (but I do understand that I have currently tested one format and the CA community can have many more, so always tricky).

collectiveaccess:
Yes exactly.

collectiveaccess:
Just as an addendum, looking at this more closely the only real reason to force a separator in the duplication process is for non-SERIAL numbers. In those cases suffixes must be appended to maintain uniqueness, and if there's no separator it can be difficult for the user to know a suffix was added. Additional code has been committed to maintain this behavior.

Thank you for surfacing and resolving this issue!

peterjanssens:
thx!

PR #1332, opened 2022-03-31
Implement merge mode for container on Prepopulate plugin

Mystikal91:
Merge mode was present in documentation but was not working (and, to me, it looks like it was never implemented, at least not in this case). This changes should fix that

Please, double check the changes, I'm not sure if it will break some users configuration (merge was working as overwrite before this changes). As far as I can test it, it should work fine

It's also my first commit ever on github, so triple check it


collectiveaccess:
Looks good to me. Thank you!

PR #1334, opened 2022-04-15
Add 5 options to apply-prepopulate-rules

Mystikal91:
Thanks to the recent change where apply-prepopulate-rules can take options from the command line, I've added 5 new options:

- restrictToTables/excludeTables: used to restrict or exclude tables
- restrictToRules/excludeRules: used to restrict or exclude rules
- findQuery: it's passed as an argument for find() function. It's an associative array formatted as json. I will try to better document this parameter on the manual


Mystikal91:
I've also added a commit to prevent a PHP Warning caused by an array_shift inside a foreach loop condition

PR #1338, opened 2022-04-25
remove child navigation button in hierarchy browser for childless listitems

peterjanssens:
Hi,

The code was already modified before but can't the extra testing for level 0 be dropped altogether. It doesn't seem user-friendly/intuitive to have the arrow navigation for children when there are no child items to be expected (the first test). 
You still can select the item, and there doesn't seem to be other functionality to the childless button (other than clearing the (previous) next level) - or I may have missed that feature :p


collectiveaccess:
There was a reason for this long ago, but I cannot remember what it was :-/ 
I think your change is a good idea.

PR #1341, opened 2022-05-11
Fix broken URL in darwincore.xml

miqrogroove:
The infoUrl value is broken and needs to be updated or removed.


collectiveaccess:
Thanks for this. The URL you've changed it to will break soon, as we're updating the documentation. But we'll migrate it when the time comes.

PR #1350, opened 2022-06-27
Include fix for tileviewer scrolling in master

miqrogroove:
All other parts of this system use a parabolic coefficient when calculating the relative position. It was missing here and simply needed to be added.


miqrogroove:
For history, see https://github.com/collectiveaccess/pawtucket2/issues/35

PR #1362, opened 2022-08-16
Fix language code for French translation

tacman:


collectiveaccess:
This profile has been deleted at the request of the copyright holder. No modifications can be accepted.

PR #1363, opened 2022-08-23
account for decimal separator while trailing zero cleaning

peterjanssens:
there is some cleaning done in Weight attribute for trailing zero's but that takes point as decimal separator hard coded, while some locales have comma as the decimal separator and that can give unwanted results.

For instance 1.160 is "cleaned" to 1.16 where is should stay as-is (1160,00 or 1.160 without decimals)

The registered value is always correct in the db (value_decimal), it's just this presentation that can be a bit off for some values (1.161 stays 1.161 because it doesn't go through the cleaning)


collectiveaccess:
Excellent, thanks!

PR #1367, opened 2022-09-16
Fix search query builder fulltext search

kehh:
In older versions of Providence it was possible to perform searches such as `_fulltext:foobar` via the API or as a part of search query syntax and in the search query builder.

This restores that ability.

### DIFF
--- a/app/lib/Plugins/SearchEngine/SqlSearch2.php
+++ b/app/lib/Plugins/SearchEngine/SqlSearch2.php
@@ -382,10 +382,10 @@ class WLPlugSearchEngineSqlSearch2 extends BaseSearchPlugin implements IWLPlugSe
 	 	}
 	 	$ap = $field ? $this->_getElementIDForAccessPoint($subject_tablenum, $field) : null;
 	 	$words = [$term->text];
-		if($field && $ap && !is_array($ap)) {
-			array_unshift($words, $field);
-			$field = null;
-		}
+	 	if($field && !is_array($ap)) {
+	 		array_unshift($words, $field);
+	 		$field = null;
+	 	}
 	 	$indexing_options = caGetOption('indexing_options', $ap, null);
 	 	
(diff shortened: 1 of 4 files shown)


app/lib/Plugins/SearchEngine/SqlSearch2.php
```diff
@@ -382,10 +382,10 @@ private function _processQueryTerm(int $subject_tablenum, $query) {
 	 	}
 	 	$ap = $field ? $this->_getElementIDForAccessPoint($subject_tablenum, $field) : null;
 	 	$words = [$term->text];
-	 	if($field && !is_array($ap)) {
-	 		array_unshift($words, $field);
-	 		$field = null;
-	 	}
+		if($field && $ap && !is_array($ap)) {
```
collectiveaccess, line 385: _getElementIDForAccessPoint() returns null or an array, so we're effectively removing the ability to treat word1:word2 values as text when word1 is not a valid access point here. This behavior is actually needed, so if we want to support _fulltext as an access point we'll need to explicitly test for it.

kehh: OK so something like: 
```php
<?php
		if($field && $field !== '_fulltext' && !is_array($ap)) {
			array_unshift($words, $field);
			$field = null;
		}
```

collectiveaccess:
Or add code to treat _fulltext as a valid AP in _getElementIDForAccessPoint()

kehh:
> Or add code to treat _fulltext as a valid AP in _getElementIDForAccessPoint()

OK I've tried to do it in that method but come unstuck with all the different return points. Any pointers as to what I should return as an access point definition?

kehh:
OK @collectiveaccess I've adapted this so that when performing a `_fulltext` search it does not send the prefix with the search query. Queries still parse correctly

PR #1370, opened 2022-10-25
Only add a single related item when always quick adding in history tracking

kehh:
* Previously a quick add would create two inputs with the related movement/loan/occurrence

Before: 
image

After:
image


collectiveaccess:
Thank you! Looks good.

PR #1372, opened 2022-11-03
Add composer packaging.

kehh:


collectiveaccess:
Good idea :-) Thanks.

PR #1376, opened 2022-12-20
fix user_volatile_vars array corrupted by _changed boolean

peterjanssens:
hi,

We got a fatal error Uncaught Error: Cannot unset offset in a non-array variable where there is attempt to unset on opa_volatile_user_vars_have_changed (the boolean) and that probably needs to be opa_volatile_user_vars (the array)??

Added more verbose (array/existingkey) test-before-unset while being there, and the neighbouring unset as well.


collectiveaccess:
`unset($this->opa_volatile_user_vars_have_changed[$ps_key]);` needs to be removed. I'll commit that change shortly.

PR #1377, opened 2022-12-20
fix Uncaught TypeError: Cannot access offset of type string on string

peterjanssens:
Hi,

We got this Uncaught TypeError: Cannot access offset of type string on string on the relationshipGenerator plugin

$va_trigger[$key], looks indeed that the current trigger (looped from $va_triggers) is already a string, and cannot be read with index key, so without knowing all context keeping the original $va_triggers seems the only choice here ? (makes you wonder how forgiving php<8 was if that wasn't a problem earlier - dates 2014)

So not completely sure this will "solve" the problem, but it clears the obvious typeError.

Peter


collectiveaccess:
No idea. This was contributed code. I'm going to accept your PR, but we're just going to remove the plugin from the distribution. If anyone still uses them they can test/support/update it.

PR #1380, opened 2022-12-20
consistently null coalescing on aggregateUnique

peterjanssens:
consistent null coalescing testing on aggregateUnique (warning on non-existance in DisplayTemplateParser)


collectiveaccess:
Thank you!

PR #1381, opened 2022-12-21
no postError field existance on non-fields

peterjanssens:
when debug_output enabled we get postErrors on non-errors.

in this case going from a set to the batch editor the BaseModel is loaded for instance with pm_id array(2) { [21]=> int(28) ["__all__"]=> int(14) }

The condition here is tested for vs_field 21 that is not a field. It is probably 'correct' that is not a field, but not important feedback either ?

I'm guessing here that the routine can be exited too, so kept the return (false) as-is

### DIFF
--- a/app/lib/BaseModel.php
+++ b/app/lib/BaseModel.php
@@ -2070,7 +2070,7 @@ class BaseModel extends BaseObject {
 				}
 
 				if (!$this->hasField($vs_field)) {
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Hmm, there's more to this I think. I'll investigate.

peterjanssens:
14 is the identifier for the (standard) editor ui, not sure yet where the 21 points to

PR #1391, opened 2023-02-12
Add php-http/discovery to composer allow-plugins

goetzk:
From Composer 2.2.0 there is a mechanism to permit certain plugins to execute code during a composer run.
https://getcomposer.org/doc/06-config.md#allow-plugins

This change gives php-http/discovery that permission to stop composer stopping mid run to ask for permission.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,8 @@
     "scholarslab/bagit": "~0.2"
   },
   "config": {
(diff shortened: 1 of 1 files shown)


goetzk:
1456bb279ecb48efa0c836111b6f75919508e152 fixed this, I just hadn't found it yet.

Issue #1392, opened 2023-02-12
Fix permission script does not change media root directory


goetzk:
Hi,
While trying to perform an install we found that in 1a8aac42975974d67f7dcd66abb5913aec475604 `fix-permissions` was changed to operate on a sub directory of media not the top level directory.
```
@@ -535,7 +527,8 @@
                                chmod($vs_path, 0770);
                        }
                        if (!$po_opts->getOption("quiet")) { CLIUtils::addMessage(_t("Fixing permissions for the media directory (media) for ownership by \"%1\"...", $vs_user)); }
-                       $va_files = caGetDirectoryContentsAsList($vs_path = __CA_BASE_DIR__.'/media', true, false, false, true);
+                       $media_root = $config->get("ca_media_root_dir");
+                       $va_files = caGetDirectoryContentsAsList($media_root, true, true, false, true, ['includeRoot' => true]);
 
                        foreach($va_files as $vs_path) {
                                chown($vs_path, $vs_user);
```
Combined with `app/conf/global.conf ` which has this:
```
ca_media_root_dir = <ca_base_dir>/media/<app_name>
```

While checking the expected behaviour I found the manual (correctly) warns about setting the media directories permissions on its install page (https://manual.collectiveaccess.org/setup/Installation.html) along with the other locations which are changed by fix-permissions.

I'm hoping to clarify if this change to the media path is intended, and if yes would it introduce problems to also change the parent media directory's permissions and or ownership too?

collectiveaccess:
It probably would not be a problem to change the parent directory's permissions, as the only contents _should_ be the application-specific media directory. The change was made, I believe, to limit permissions errors to those directories that are absolutely necessary. 

Is this behavior causing a problem for you?

goetzk:
This problem occurs on a fresh install if `media/` is not writeable by the web server as it prevents creation of `media/appname` . Running fix-permissions seems the obvious solution (and #1393 made it look like the directory would change) but while fix-permissions fixes all the other directories required for CA to install it doesn't do this one.

collectiveaccess:
Ah yes I see. We should change it then.

PR #1393, opened 2023-02-12
Adjustments to fixing permissions messages

goetzk:
Several of the messages produced by fix-permissions are confusing, updating the change paths to improve accuracy.


goetzk:
I believe this change stands irrespective of what happens in #1392 .

PR #1394, opened 2023-02-14
update iconclass API lookup

peterjanssens:
The iconclass service old API endpoint seems to redirect to an html landingpage now, ie.:

http://iconclass.org/rkd/9/?q=europa&q_s=1&fmt=json

and no longer to a json with records/results.

I tried to stay as close to the original implementation with the (new?) documented API's https://iconclass.org/docs

What appeared to have worked in the past in one request, is now to be done in subsequent requests, the first (language based) search to get results with key, the next to get the (language based) labels for the resulting keys/ids.

https://iconclass.org/api/search?q=europa&lang=en&size=999&page=1&sort=rank&keys=0
[
https://iconclass.org/92B1218.json,
..
]

It works on both php7 as php8 installations.

PS
I've kept the implementation simple, but I would like to propose a variation in a later PR(s).
For some InformationServices the feedback/suggestions you get can be overwelming to select from (like picking the correct Mozart if you have a dozen WA Mozarts). The jQuery autocompletion allows both 'label' and 'value', and we currently only use one of them, but it would be nicer to have some useful extra metadata alongside the (dropdown)label to aid the selection, yet keep the value(label) simple and concise in the end. I've done it for another IService we cannot disclose yet and it didn't need many changes without loosing compatibility to existing services.

### DIFF
--- a/app/lib/Plugins/InformationService/Iconclass.php
+++ b/app/lib/Plugins/InformationService/Iconclass.php
@@ -86,20 +86,25 @@ class WLPlugInformationServiceIconclass Extends BaseInformationServicePlugin Imp
 		} else {
 			$vs_lang = 'en';
 		}
-		
-		
-		$vs_content = caQueryExternalWebservice(
-            $vs_url = 'http://iconclass.org/rkd/9/?q='.urlencode($ps_search).'&q_s=1&fmt=json'
-		);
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Thanks

PR #1398, opened 2023-02-23
PHP8 compatibility each() fix

Monica-Wood:
PHP8 has removed the function each(), this has replaced where possible the while(list() = each()) loop with a foreach loop


collectiveaccess:
Thanks for catching this.

PR #1401, opened 2023-03-02
update key setting for jwt decode

peterjanssens:
was playing with GraphQL and got "\"kid\" empty, unable to lookup correct key" from the Firebase php-jwt, where getKey expects a jwt Key or an array of Keys with kid; but it currently got a string, not a key. Apparently the signature got changed where decode gets the Key directly.

PS: documentation wasn't clear (or no longer valid) where the jwt token should go, I think I used to pass it as a query parameter but now it has to be passed as an argument in the body. Maybe it got there automagically in earlier php7 ?
Passing via Authorization Bearer {token} Header still works.

`{"query":"query {get(jwt:\"eyJ0eXA..., table:'ca_objects', identifier:...`

### DIFF
--- a/app/lib/Controller/Request/Session.php
+++ b/app/lib/Controller/Request/Session.php
@@ -384,7 +384,7 @@ class Session {
 	 */
 	public static function decodeJWT(string $jwt, string $key) {
 		if (!$key) { $key = Configuration::load()->get('jwt_token_key'); }
-		return JWT::decode($jwt, $key, ['HS256']);
+		return JWT::decode($jwt, new Firebase\JWT\Key($key, 'HS256'));
 	}
 	# ----------------------------------------
 }


collectiveaccess:
I guess this is a PHP 8 specific change? It's not throwing errors on PHP7.

collectiveaccess:
Also JWT does not need to be passed passed as a parameter in the query, although it could be if you wish (we supported it for testing early on, but really it's unnecessary). 

Typically you'd pass it as a Bearer token in an Authorization header.

peterjanssens:
Bearer token in header is ok; I picked up the query param option from the existing documentation (https://manual.collectiveaccess.org/developer/web_api/graphql/auth.html#jwt-configuration-options) in my earliest testing.

It was when the queryparam no longer worked that I got inspired by the self::authenticate($args['jwt']) in the AuthController to pass it in the query itself; and it ..  _worked_ :) at least for the Item endpoint, but not for some other ones I tried (Schema,..)

(maybe better drop the queryparam option from the docs? Is there a way we can participate on the docs .. some day ?)

collectiveaccess:
The JWT query parameter should work for all endpoints. I'll check the schema endpoint and see. While we could drop it, I suppose it's still useful in some contexts. 

Regarding docs, you can fork the docs and submit changes via pull requests.

PR #1402, opened 2023-03-07
editing default profile

gautiermichelin:
Avoid an error if geonames is not filled up in the Relations screen "Geonames : was blank".

Capture d’écran 2023-03-07 à 09 43 14


collectiveaccess:
Why not just disable the "requireValue" setting on the element?

PR #1407, opened 2023-03-16
correct count on browse summaries

peterjanssens:
when there are many ACL rules, and more than one fits your user and group memberships, count can be higher than it should (ie. equal to the the filtered results) because the acl joins duplicates rows.
Not a fan of distinct but I think it is appropriate here.

(item_id could be part of that values list, but I think that value -like a listitem- is always duplicated in longtext1 ?)

### DIFF
--- a/app/lib/Browse/BrowseEngine.php
+++ b/app/lib/Browse/BrowseEngine.php
@@ -4338,7 +4338,7 @@
 					        $va_params[] = $va_container_ids[$vs_container_code];
 					    }
 						$vs_sql = "
-							SELECT COUNT(*) as _count, ca_attribute_values.value_longtext1, ca_attribute_values.value_decimal1, ca_attribute_values.value_longtext2, ca_attribute_values.value_integer1, ca_attribute_values.element_id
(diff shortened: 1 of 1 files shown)


collectiveaccess:
You're using ACL in your instance? I see the issue and will accept the PR, but am wondering if you've noticed anything else related to ACL. We haven't done a lot of testing with that in a while.

peterjanssens:
yes we use the ACL's in some of our installations.
The count error was usually small (and sneaky :p) and for especially to some users that are cross groups/roles, but you can easily get the number to explode when you add more relevant groups. I understand that the logic tries to find the items you are allowed to see, but when you match on multiple rules for certain items they (objects,..) should be considered individually. 
I guess the same scenario holds for all the main constructs (ca_objects, ca_entities,..); I was looking at it through object browse, so I took the attributes.row_id (instead of object_id) to allow compatibility for the rest.

peterjanssens:
Or maybe I haven't understood your follow-up question correctly: did we notice anything ACL related in the scope of the new php8 version? We have only just started testing really, we will give it some extra attention then.

PR #1423, opened 2023-05-31
NOJIRA : hookRenderSideNav to allow left menu (screens list) custo

gautiermichelin:
This hook allows screen list customization : I've built lately a very simple plugin to allow reverse views of tables used inside metadata (ObjectsAttributeValue, EntitiesAttributeValue...). This plugin allow back and forth navigation btw entities used inside a container and the object its in.
To be able to plug it on the screens list, I've made this simply patch, that could be used by others.

Name is meant to be fastly understandable, params are self-understandable.


gautiermichelin:
Hi @collectiveaccess this is a very simple addition to add a hook for screens inside UIs. Could you have a look ?

PR #1429, opened 2023-06-08
Halt task queue if handler not found.

bri-gaia:
Made this more consistent with other scenarios where the "queue is halted"


collectiveaccess:
Thanks!

PR #1437, opened 2023-06-19
Update MultipartIDNumber - getSortableValue() 'PARENT' functionality …

bri-gaia:
…to only explode on separator if it exists.


bri-gaia:
Shouldn't have made an PR for this yet, needs more testing on my end

PR #1443, opened 2023-06-26
Update Set deletion query for improved performance.

bri-gaia:


app/models/ca_sets.php
```diff
@@ -396,13 +396,13 @@ public function delete($pb_delete_related=false, $pa_options=null, $pa_fields=nu
 
 			// quickly delete attribute values
 			$this->getDb()->query('
-				DELETE FROM ca_attribute_values WHERE attribute_id IN
-				(SELECT attribute_id FROM ca_attributes WHERE table_num=? and row_id IN (SELECT item_id FROM ca_set_items WHERE set_id = ?))
+				DELETE av.* FROM ca_attribute_values av JOIN ca_attributes a ON av.attribute_id = a.attribute_id
+				JOIN ca_set_items s on a.row_id = s.item_id  WHERE a.table_num=? and s.set_id = ?
```
collectiveaccess, line 400: If we're going to use a multi-table delete then perhaps we should delete both ca_attributes and ca_attribute_values in a single query?

bri-gaia: Makes sense. On it

bri-gaia: @collectiveaccess ah, no we cannot do this in one statement because of the foreign key constraint. (at least not without a workaround that I don't feel comfortable doing.)

PR #1449, opened 2023-06-26
Add ability for plugins to define custom expressions …

bri-gaia:
…for use in display templates.


kehh:
Very similar code to this has been in use in production for the past couple of years. Similar to #1450 this moves this functionality into a plugin. The plugin can then register custom functions that can be called:

```php
    public function hookRegisterExpressions($pa_options) {
        $fancy_methods = [
            'fancyfyTheText',
        ];
        foreach ($fancy_methods as $method) {
            $custom = function(...$args) use ($method) {
                if (is_callable("FancyClass::{$method}")){
                    $ret = fancyClass::$method(...$args);
                } else {
                    throw new Exception("Method fancyClass::$method does not exist and cannot be called.");
                }
                return $ret;
            };
            $pa_options['expressions']['FancyText' . $method] = xcallable($custom);
        }

        return $pa_options;
    }
```

PR #1450, opened 2023-06-26
Implement hook to allow components to define their own display components.

bri-gaia:


kehh:
This makes it possible to put the contents of a bundle into version control (in a plugin). Then to consume it you just add the placement to a display. Very similar code to this has been in use in production for a couple of years, but this makes it possible to do this through a plugin.

PR #1455, opened 2023-06-29
Provcau 99 disable hierarchical indexing

bri-gaia:


kehh:
@bri-gaia I think this is the wrong fork ;) (and too much content)

bri-gaia:
Thanks!

PR #1462, opened 2023-07-10
Implement AJAX Summary Displays.

bri-gaia:
Add option to render summary displays using AJAX.


collectiveaccess:
This is PR with changes in 23 files. Can you provide something more than a one line description of what you're trying to accomplish here, and why it is necessary?

bri-gaia:
Some of the data we are displaying in the summary tab is quite heavy, so we implemented AJAX loading so that the page would render immediately and we could show a loading graphic.

By individually loading in the data elements the smaller requests can be populated without being held up by the more intensive queries.

I have requested that my tech lead provide additional context to this PR if possible.

kehh:
@collectiveaccess what @bri-gaia has put above is pretty much the gist of it. Essentially very complex templates can take some time to render especially if they are traversing many relationships / aggregating values from all descendants. This allows the page to render without being held up by slower templates.

kehh:
There's a switch for this functionality in app.conf - that could be defaulted to off if you prefer?

bri-gaia:
Additional info:

* The crux of the change of this is implemented in `BaseEditorController.php` refactoring `Summary()` into separate functions that can be called individually if configured to us AJAX while maintaining the current functionality if AJAX is not enabled. 
* I have implemented `renderGeneric()` and updated the call to `template_test_html.php` as the relative path only worked for summary's under the `themes/default/views/editor` directory, causing issues with, for example, the list_editor.
* The changes to the 17 `summary_html.php` files should all be identical, and I have attempted to modify as little in those files as possible to make that visible. 

Apologies for the massive PR.

PR #1467, opened 2023-07-14
Dev/php8 - PHP8 Fixes

Monica-Wood:
Small PHP8 Fixes

### DIFF
--- a/app/lib/BatchProcessor.php
+++ b/app/lib/BatchProcessor.php
@@ -1469,15 +1469,15 @@ class BatchProcessor {
 		
 		if (isset($pa_options['sendMail']) && $pa_options['sendMail']) {
 			if ($vs_email = trim($po_request->user->get('email'))) {
-				$info = $t_importer?->getInfoForLastImport();
+				$t_importer ? $info = $t_importer->getInfoForLastImport() : null;
 				caSendMessageUsingView($po_request, array($vs_email => $po_request->user->get('fname').' '.$po_request->user->get('lname')), __CA_ADMIN_EMAIL__, _t('[%1] Batch metadata import completed', $po_request->config->get('app_display_name')), 'batch_metadata_import_completed.tpl', 
 					[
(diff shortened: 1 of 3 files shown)


app/lib/BatchProcessor.php
```diff
@@ -1469,7 +1469,7 @@ public static function importMetadata($po_request, $ps_source, $ps_importer, $ps
 		
 		if (isset($pa_options['sendMail']) && $pa_options['sendMail']) {
 			if ($vs_email = trim($po_request->user->get('email'))) {
-				$info = $t_importer->getInfoForLastImport();
+				$info = $t_importer?->getInfoForLastImport();
```
collectiveaccess, line 1472: What are you trying to do here?

Monica-Wood: Getting this error when calling process-task-queue with the -r flag, when there are stuck processes to restart.

```
PHP Fatal error:  Uncaught Error: Call to a member function getInfoForLastImport() on null in /srv/providence/app/lib/BatchProcessor.php:1472
Stack trace:
#0 /srv/providence/app/lib/Plugins/TaskQueueHandlers/metadataImport.php(106): BatchProcessor::importMetadata()
#1 /srv/providence/app/lib/TaskQueue.php(328): WLPlugTaskQueueHandlermetadataImport->process()
#2 /srv/providence/app/lib/Utils/CLIUtils/Cron.php(46): TaskQueue->processQueue()
#3 /srv/providence/support/bin/caUtils(172): CLIUtils::process_task_queue()
#4 {main}
  thrown in /srv/providence/app/lib/BatchProcessor.php on line 1472
```

I am not sure if $t_importer is meant to be not null in this situation, but this allows the getInfoForLastImport to be called without an error even if $t_importer is null

Monica-Wood: Updated in new commit.  I have tested that this still sends emails correctly with a normal metadata import.

PR #1469, opened 2023-08-02
Inspector headline show only 'viewing' if only one record type

Monica-Wood:
I noticed in one if my systems that object records were showing 'viewing .... ' where collections where showing 'editing ...'  I tracked it down to this line that would only pick 'editing ...' if there was more than 1 type.    Not sure if there is a reason for this to be 1, but changing it to 0 fixed the issue.


collectiveaccess:
Fixed with somewhat different approach.

PR #1485, opened 2023-08-22
Allow configuration for quickly searching based on idno regex

kehh:
* Fix undefined variable which was preventing this from working
* Retrieve $t_table earlier in the method

### DIFF
--- a/app/lib/Search/SearchEngine.php
+++ b/app/lib/Search/SearchEngine.php
@@ -187,24 +187,26 @@ class SearchEngine extends SearchBase {
 				}
 			}
 		}
-		
+
+		$t_table = Datamodel::getInstanceByTableName($this->ops_tablename, true);
+		$vs_idno_fld = $t_table->getProperty('ID_NUMBERING_ID_FIELD');
+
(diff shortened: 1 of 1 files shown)


app/lib/Search/SearchEngine.php
```diff
@@ -187,24 +187,26 @@ public function doSearch($ps_search, $po_result=null, $pa_options=null) {
 				}
 			}
 		}
-		
+
+		$t_table = Datamodel::getInstanceByTableName($this->ops_tablename, true);
+		$vs_idno_fld = $t_table->getProperty('ID_NUMBERING_ID_FIELD');
+
         if ((is_array($va_idno_regexs = $this->opo_search_config->get('idno_regexes'))) && (!preg_match("/".$this->ops_tablename.".{$vs_idno_fld}/", $ps_search))) {
```
kehh, line 194: before this change $vs_idno_fld was not defined here

kehh:
Note that after looking at this, it appears this functionality still works, but the first check whether the search is already for an idno fails. But this change avoids the undefined variable.

collectiveaccess:
Thank you.

PR #1488, opened 2023-08-25
Fix fatal error in pdo_mysql

kehh:
I experimented with pdo_mysql and found this fatal error.


collectiveaccess:
There has been no testing with pdo in a long while. And none with PHP 8. If someone/anyone can take that on it would be greatly appreciated.

PR #1500, opened 2023-09-13
Update docx_results.php : missing field names in the docx export

gautiermichelin:
Update docx_results.php : missing field names in the docx export


gautiermichelin:
Nice and fast, thanks !

PR #1534, opened 2024-02-06
Add rank to site pages

Monica-Wood:
Enable the ability to sort menu items, or content pulled into other pages.

### DIFF
--- a/app/models/ca_site_pages.php
+++ b/app/models/ca_site_pages.php
@@ -82,6 +82,13 @@ BaseModel::$s_ca_models_definitions['ca_site_pages'] = array(
 				'DISPLAY_FIELD' => array('ca_locales.name', 'ca_locales.language', 'ca_locales.country'),
 				'LABEL' => _t('Locale'), 'DESCRIPTION' => _t('Locale of page'),
 		),
+		'rank' => array(
+				'FIELD_TYPE' => FT_NUMBER, 'DISPLAY_TYPE' => DT_FIELD,
+				'DISPLAY_WIDTH' => 10, 'DISPLAY_HEIGHT' => 1,
+				'IS_NULL' => false,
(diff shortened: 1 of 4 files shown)


collectiveaccess:
What exactly is this needed for?

Monica-Wood:
So you can determine the order that pages show up.  Currently you can only show them based on the the order they are created.  If you create a new one and need it to show between 2 existing, then you can't set this.
Examples are generated menus, or in my case I am generating cards on the front page (and I will also create index pages for other page types) with media and blurb that link off to the full content page.

collectiveaccess:
The pages are accessed by URL on the front-end. There is no order. I guess you're using these on the back-end in a menu?

Monica-Wood:
I am using php code to collect all site pages of a certain template and outputing the media and first bit of the page content into a bootstrap card. This then links off the page itself.   But I cannot control the exact order these pages are returned, this allows to put a 'sort' => 'rank' into the find call.  

Here is the top snippet from my code I am using in Pawtucket on the front_page_html.php

```
$t_site_templates = new ca_site_templates(array("template_code" => "About_Collection_Page"));
if($vn_collection_template_id = $t_site_templates->get("ca_site_templates.template_id")){
        $q_collections = ca_site_pages::find(array('template_id' => $vn_collection_template_id), array('returnAs' => 'searchResult', 'sort' => 'rank', 'checkAccess' => $va_access_values));
        if($q_collections->numHits() > 0){
	        $va_collections = array();
	        while($q_collections->nextHit()){
```

It can also be used to generate menus that link off to site pages created, so admins don't have to add new menu links every time a page is made.  It allows for dynamic creation of content.

Monica-Wood:
Is there anything you would like me to change to have this accepted?

Monica-Wood:
Thank you

PR #1545, opened 2024-03-11
Pawtucket spam

bri-gaia:
@kehh we briefly talked about getting rid of this spam. 

This approach will only print process logs if there is a value. I'm not sure on the context of this, so not sure what the best approach is.

Is this being used? Can we disable the plugin entirely? or perhaps the Periodic task? If this isn't doing something we need, it might help with performance to completely remove/disable it.
Do we want a config option to suppress these? do we want to suppress even if there is a value?


bri-gaia:
Wrong target branch sorry.

PR #1562, opened 2024-04-12
Fix for "Download all" from dev/php8 in 1.8rc5

amergelCGI:


Monica-Wood:
Identical to this commit https://github.com/collectiveaccess/providence/commit/cda042ef8dab50819442af7a3733ac31a13b5cdb

PR #1585, opened 2024-05-22
Improve error checking for Shib field map

goetzk:
Previously error checking was almost non existant and performed its task completely silently.
This change brings missing elements in the map right to the fore by refusing to continue while uid and email are missing.

Rounding out the change is adding a stacktrace to "Could not create SimpleSAML auth object", we've been carrying this change locally for some time and its a great help in debugging.

### DIFF
--- a/app/conf/authentication.conf
+++ b/app/conf/authentication.conf
@@ -282,7 +282,8 @@ shibboleth_service_provider = default-sp
 # Name of Shibboleth token cookie
 shibboleth_token_cookie = shib_test
 
-# Mapping of CA user fields to Shibboleth attributes. Required fields are marked
+# Mapping of CA user fields (on the left) to Shibboleth attributes (on the right).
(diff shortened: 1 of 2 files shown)


collectiveaccess:
I already have some pending changes that add detailed debugging logging. I'm also flipping the map, as we have users who reuse the same attribute for multiple values (Eg. email for both email and user_name), and the current implementation doesn't allow it. I'll merge your suggestions into as make sense and then you can give it a try.

collectiveaccess:
Just pushed a `dev/shibboleth-logging` with changes. Can you give it a try? There's a new `shibboleth_debug` option in authentication.conf to trigger debug logging,

goetzk:
I've only lightly tested but it appears to be behaving as expected - thanks.
I will close this in deference to the (recently merged) 9c1e4aa30b4d9dfa30eaed57ce855a580441f74e

PR #1587, opened 2024-06-18
Make HTMLPurifier path configurable

goetzk:
This removes the only writeable path under vendor/, allowing the full directory path to be marked read only

To enable the change a new subdirectory is created in tmp to hold serialisers working data.

### DIFF
--- a/app/conf/app.conf
+++ b/app/conf/app.conf
@@ -3127,6 +3127,10 @@ purify_all_text_input = 1
 purify_allow_external_references = 0
 
 
+# Override path to HTMLPurifier working directory.
+# Set to null to disable caching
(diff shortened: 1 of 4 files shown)


goetzk:
I should note I've 'run tested' this but the system in question doesn't appear to have updated the serialiser files for a while so I'm not completely certain this is bug free.

goetzk:
I've now tested this again on a new system. There was at least one place during the configuration test which wasn't previously caught in the PR, hopefully thats now resolved.

goetzk:
I just discovered there is an upstream issue about this at https://github.com/ezyang/htmlpurifier/issues/71 so crosslinking here.

goetzk:
Hi @collectiveaccess , is there anything you need from my side to review this PR?

PR #1590, opened 2024-06-27
NOJIRA : error 500 in collection advanced search

gautiermichelin:
This line is missing, it's required, and found by analyzing SearchObjectsAdvancedController.php that is OK.


app/controllers/find/SearchCollectionsAdvancedController.php
```diff
@@ -68,6 +68,7 @@ public function __construct(&$po_request, &$po_response, $pa_view_paths=null) {
 			);
 		}
 		$this->opa_views = caApplyFindViewUserRestrictions($po_request->getUser(), 'ca_collections', ['returnAll' => $this->request->config->get('enable_full_thumbnail_result_views_for_ca_collections_search'), 'type_id' => $this->opn_type_restriction_id]);
+		$this->opo_browse = new ObjectBrowse($this->opo_result_context->getParameter('browse_id'), 'providence');
```
collectiveaccess, line 71: Why would you add an object browse in a collections search?

PR #1593, opened 2024-07-10
add addScript to MetaTagHandler

peterjanssens:
This adds the addScript function to the MetaTagHandler, with rendering of the script in <meta>.

This would allow to inject (js) scripts specific for plugins (example: /app/plugins/myCustomPlugin/assets/js/my_custom_script.js), and would allow easier handling (ie. via enabling/disabling plugin and not by tampering /conf/local/assets.conf)

it follows the other meta elements (like stylesheet links)
MetaTagManager::addScript(__CA_URL_ROOT__."/app/plugins/myCustomPlugin/assets/js/my_custom_script.js",'text/javascript');

### DIFF
--- a/app/lib/MetaTagManager.php
+++ b/app/lib/MetaTagManager.php
@@ -90,6 +90,24 @@ class MetaTagManager {
 			'type' => $ps_type
 		);
 		
+			return true;
+		}
+		# --------------------------------------------------------------------------------
+		/**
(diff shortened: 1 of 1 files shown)


bruceklotz:
Looks like there a typo - ~line 155 - Missing '}' to close out getHTML function

collectiveaccess:
Fixed this.

PR #1596, opened 2024-07-23
Readded: Small Fix for mediaupload problems caused by missing metadata

CGitPrinz:
This was a hotfix we needed for our big rollout. There is a better solution, but it requires to rewrite some of the logic in this function (app/helpers/gisHelpers.php: caGISminutesToSignedDecimal)

The upload process of media/files works fine, but when clicking on "Save", we got errors because of missing metadata from those specific images, that are probably needed


goetzk:
@CGitPrinz you are trying to merge to dev/php8 from a fork of master. you'll need to rebase/retarget your PR so the CA team can merge it.

That said, a quick look at the commits in this changeset suggests https://github.com/collectiveaccess/providence/pull/1596/commits/c6e86b7af16dfa2592461aa46c61e3cfe9f0a4a8 is the only change so they may be able to cherry pick that one directly from your repo if they're willing.

PR #1615, opened 2024-09-04
NOJIRA BatchProcessor : handles hookSaveItem after a batch processing

gautiermichelin:
hookSaveItem is fired after a Save event in the interface, allowing to prepopulate metadatas for example. If you batch process records on those metadatas, the subsequent computes are not executed.

This PR aims to fire hookSaveItem after a batch treatment.


gautiermichelin:
hello @collectiveaccess this still to be included in the current version. Would you be ok to have a look at ?
Best to all of you for the end of the year !

gautiermichelin:
Thanks

PR #1650, opened 2024-11-12
NOJIRA : add a 3rd attempt for failsafe video thumb extraction

gautiermichelin:
NOJIRA :
- correction of the 2nd : "// try again, with -ss 1" => the -ss 1 wasn't applied
- add a 3rd failsafe video thumb extraction, without the "pad" indication inside the -vf, this secured me many cases where 1st and 2nd attempts gave nothing


collectiveaccess:
What version(s) of ffmpeg is this tested with?

gautiermichelin:
bump ?

collectiveaccess:
This is against 1.7. Is that what you want?

PR #1652, opened 2024-11-22
bug always the same random object on basemodel

behnaz-zarriz:
our server is up to date, but the random object selection method (SELECT CEIL(RAND()) is not working.

This query that I modified now works because it first selects random IDs using ORDER BY RAND() and then retrieves the full records using an INNER JOIN

### DIFF
--- a/app/lib/BaseModel.php
+++ b/app/lib/BaseModel.php
@@ -11934,15 +11934,17 @@ $pa_options["display_form_field_tips"] = true;
 		}
 		
 		$vs_sql = "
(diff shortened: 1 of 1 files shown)


collectiveaccess:
What version of MySQL are you running?

behnaz-zarriz:
> What version of MySQL are you running?

Version 8.0.36

collectiveaccess:
Performance for ORDER BY RAND() falls off a cliff for large data sets, which is why this current query is so wacky. We'll need to find another way.

PR #1655, opened 2024-11-28
Customisable temp path

goetzk:
Move to replace many and varied ways to access the CA app/tmp path with a single configurable variable.
This can be set via setup.php along with other site specific variables allowing a separation between state per site.


goetzk:
I forgot to mention - this has not been tested yet.

goetzk:
@collectiveaccess I realised this has been based on 2.0rc1 instead of dev/php8; would you like me to rebase and re propose against dev/php8?

goetzk:
I have moved this back to dev/php8.

goetzk:
I've been testing this and it appears `__CA_TEMP_DIR__` is being correctly defined in setup.php and `ConfigurationCheck.php` tests like 'is temp writeable' are succeeding.

my attempt to feed the setting in to configuration files though, is failing.
```
root@multitest:/srv/providence_php82# rgrep ca_temp_dir app/
app/conf/app.conf:taskqueue_tmp_directory = <ca_temp_dir>
app/conf/app.conf:# Clear stray files in the ca_temp_dir directory older than specified age (in seconds)
app/conf/app.conf:purify_serializer_path = <ca_temp_dir>/purifier
app/conf/app.conf:tilepic_tmpdir = <ca_temp_dir>
app/conf/global.conf:ca_temp_dir = __CA_TEMP_DIR__
```

When the `ConfigurationCheck.php` test for `$vs_purifier_path` (itself the configuration item `purify_serializer_path`) runs it fails because the prefix is missing from serializer path.

> It looks like the directory for HTML filtering caches is not writable by the webserver. Please change the permissions of /purifier and enable the user which runs the webserver to write to this directory.

goetzk:
Update: waiting 10 minutes "resolved" the error. i don't know if it was a caching issue or something else but i have a site loading with this configuration in place.

goetzk:
I have force pushed an update based on the current master; providence 2.0 codebase. I haven't yet tested this update.

goetzk:
I've just re-tested this on dev/php8 (commit head 50ceabf349f9f97b59cd87febde77a5cb3d0acd9).
- no `__CA_TEMP_DIR__` defined defaults to `app/tmp`
- with `__CA_TEMP_DIR__` defined the provided path is used.

Then tested on dev/2.0 (commit head b95fff07e566ca9290afb1e9a46b40eb7688f9ed).
- no `__CA_TEMP_DIR__` defined defaults to `app/tmp`
- with `__CA_TEMP_DIR__` defined the provided path is used.

If `__CA_TEMP_DIR__` is defined as an empty string, an empty string is used and things will break. I thought about handling that but none of the other checks in `post-setup.php` do so I assume 'trust the administrator not to mess it up' is the policy there.

PR #1682, opened 2025-01-12
A couple more typos and a couple of _t( )

yelto:


yelto:
Should I just mktrans a new messages.pot by hand and pull-request it or is there some other magic?

PR #1728, opened 2025-06-14
Revert overly broad conversion to https for LCSH

mark1ewis:
This reverts a portion of Seth's commit 2554940a77f2f3f6390b5b2c9e4486029b56acf8

The problem that commit was trying to fix is that for the Libary of Congress, the request URL must always use https now.

However the URI which identifies the type of lookup (subject headings, name authority, etc) must still use http, since that's part of the controlled identifier. The offending commit also changed the URIs.

This change restores the original URIs which were correct. Because the URI is used as the value of the vocabulary drop-down for the metadata element, this means that:
 (a) Users with pre-existing LCSH metadata elements might actually have worked, because
they have the old URI value stored until the next time they modify the metadata element
 (b) Users who created a new LCSH field after the bad commit will
probably need to re-save the metadata element after applying this fix to set the correct URI

I don't know if this is a complete fix because there are other places in the code where for example the URI is parsed out of the vocabulary value, and those might need to be updated? I'm not familiar enough with providence to know, but I suggest comparing this with the original commit to see if any other changes are required.

At least for me, without this change, the autocomplete for LCSH searches that specified a vocabulary would always fail with the latest master code. With this change (and after re-saving the LCHS metadata element to pick up the new value), autocomplete searches for LCSH vocabularies seem to work


collectiveaccess:
Thanks for this. I see the the LC vocabulary service is not returning results when an https vocabulary specifier is used. I could swear that this did work when the change was made. Regardless, I've just pushed your change + some changes to allow elements configured with the https versions to work without requiring reconfiguration.

PR #1751, opened 2025-09-04
NOJIRA : VIAF is actually broken due to changes in OCLC API, fixing

gautiermichelin:
NOJIRA : VIAF is actually broken due to changes in OCLC API, fixing it.

- removing the namespace prefix (ns1, ns2, ns3... from the array keys of the result)
- properly handling the Accept request attribute : without that, we get an error 500


gautiermichelin:
thanks

PR #1761, opened 2025-09-22
allow dynamic configuration of authentication adapter

peterjanssens:
For one of our projects we need both local users and simplesaml logins. Part of the users have access through their university IdP, part of the users can't have those accounts.
This change allows this variable to be handled through a new __CA__ const (__CA_AUTH_ADAPTER__) so that we can define this earlier on in custom setup.php. Users can select their adapter in a custom login screen (or get previously chosen adapter immediately through for instance session). Defaults to CaUsers, and existing conf/local/authentication.conf using one of the defined 'string' adapters shouldn't be affected.

We could do with just our own custom setup and custom conf/local, but maybe others could benefit from this (alternative to the allow_fallback_to_ca_users_auth)

image

### DIFF
--- a/app/conf/authentication.conf
+++ b/app/conf/authentication.conf
@@ -12,7 +12,7 @@
 #	ExternalDB : Authentication using an external database
 #   Shibboleth: Authentication via the Shibboleth SSO service
 #	HTTPHeader : User info in HTTP Headers - web server handles login
-auth_adapter = CaUsers
+auth_adapter = __CA_AUTH_ADAPTER__
 
 # User different authentication method for service calls?
 #auth_adapter_for_services = CaUsers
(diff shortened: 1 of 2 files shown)


peterjanssens:
no problem, than we keep it in our custom setup and conf/local entirely.
The use-case of mixed accounts will not be that general but it happens.

collectiveaccess:
If you want to contribute the "CA" adapter as well then it would make sense to roll this all into the distribution. I can imagine being able to choose between SAML and local login (for example) could be of interest to others.

peterjanssens:
That was our use-case indeed, some have to use one, some the other and then auth_adapter config from authentication.conf cannot be just a string (or you would need per-user file sed). It's equally manageable from the custom setup.php and the conf/local/authentication.conf but we wanted to address the (edge) case anyway.
We feed most of our consts from an .env so the custom setup is already prepared to handle special cases.

I'm not sure if I understand the 'contribute to the CA adapter' :)

collectiveaccess:
Thanks for the details. I think implementing this would be handy at some point, so I'll accept this PR with the intention that we'll add a version of what you describe in a future release.

peterjanssens:
Thanks!

PR #1766, opened 2025-10-01
Can the "Cans ...", for consistency in user_actions.conf

yelto:
Looks like the newer permission labels did not follow the previous pattern. In case it is just an unintended inconsistency, this eliminates the "Can ..." from the labels that have them.

Except... I've left the pawtucket_prefs alone, just in case they are more deliberate (some weird uppercases though).

### DIFF
--- a/app/conf/user_actions.conf
+++ b/app/conf/user_actions.conf
@@ -84,15 +84,15 @@ user_actions = {
 				description = _("Allow user to manage user defined sort field combinations")
 			},
 			can_use_plugin_tools = {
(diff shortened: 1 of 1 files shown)


yelto:
Confusing in lib/Auth/Adapters/HTTPHeaders.php: caLogEvent('LOGF', _t('Could not login user %1 using http headers are missing failed %2 [%3]', $ps_username, $vs_httpheader_username, $_SERVER['REMOTE_ADDR']), 'HTTPHeaderAuthAdapter');

PR #1853, opened 2026-02-11
filter current user from other users

peterjanssens:
when browsing sets some users are troubled by seeing their own sets listed among the "by other users" functionality.

There are a few inconsistencies still (like the 'allUsers' option key, and the navigation manage>my sets>All sets vs User Sets) but since you can have your own sets listed separately, it would make sense to have the other users sets separately (without the noise of your own sets) ?


collectiveaccess:
Or we can just relabel the option "all sets"

PR #1872, opened 2026-03-04
add logfile cleanup utility

peterjanssens:
We would like to propose a logfile cleanup utility. This allows for periodic cleanup (manual or cron) of the application logfiles (log_....txt).

Exisiting GarbageCollection utility does not handle logs, and the current logger doesn't auto delete old files.
Size limitations causes some logfiles to spill over to the root /tmp, so that directory was added

dry run is the default mode for a standard expiration of 100 days. 
days can be set to any integer >0.
mode can be set to delete for actual deletion and verbosity can be set to verbose/quiet/silent for the sake of cron logging.

### DIFF
--- a/app/lib/Utils/CLIUtils/Maintenance.php
+++ b/app/lib/Utils/CLIUtils/Maintenance.php
@@ -29,7 +29,118 @@
  * 
  * ----------------------------------------------------------------------
  */
-trait CLIUtilsMaintenance { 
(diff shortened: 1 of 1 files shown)


collectiveaccess:
You could also use `caGetDirectoryContentsAsList()`, which is a helper defined in `app/helpers/utilityHelpers.php`

PR #1890, opened 2026-04-13
HEIC to other images blocks also

Monica-Wood:


poplarmedia:
Thank you! Was about to open a ticket regarding HEIC suddenly not working after I upgraded.

PR #1901, opened 2026-04-28
Fix incorrect default for for_duplication flag preventing prepopulate execution

behnaz-zarriz:
The for_duplication flag was defaulting to true when not explicitly set in $pa_options:

    caGetOption('forDuplication', $pa_options, true)

This caused all insert/update operations to be treated as duplication,
even when no duplication was intended.

As a result, plugins (such as prepopulate) that rely on the for_duplication
flag incorrectly skipped execution.

This change updates the default value to false:

    caGetOption('forDuplication', $pa_options, false)

This ensures:
- Normal insert/update operations are not treated as duplication
- Prepopulate rules execute correctly on save
- Actual duplication behavior remains unaffected, as duplication code explicitly sets:
    ['forDuplication' => true]

This fix restores the intended behavior without impacting existing duplication workflows

### DIFF
--- a/app/lib/BundlableLabelableBaseModelWithAttributes.php
+++ b/app/lib/BundlableLabelableBaseModelWithAttributes.php
@@ -304,7 +304,7 @@ class BundlableLabelableBaseModelWithAttributes extends LabelableBaseModelWithAt
 					'table_name' => $table, 
 					'instance' => $this, 
 					'is_insert' => true, 
-					'for_duplication' => caGetOption('forDuplication', $pa_options, true)
+					'for_duplication' => caGetOption('forDuplication', $pa_options, false)
 				]
 			);
 		}
@@ -415,7 +415,7 @@ class BundlableLabelableBaseModelWithAttributes extends LabelableBaseModelWithAt
 					'table_num' => $this->tableNum(), 
 					'table_name' => $this->tableName(), 
 					'instance' => $this, 'is_insert' => false, 
-					'for_duplication' => caGetOption('forDuplication', $pa_options, true)
+					'for_duplication' => caGetOption('forDuplication', $pa_options, false)
 				]
 			);
 		}


collectiveaccess:
In what cases was prepopulate skipping execution? The code you're modifying has been in place for 5 years, and this is the first report of it not applying since then.

behnaz-zarriz:
Thanks for the question — that’s a fair point.

In our case, prepopulate was consistently skipping execution during normal save operations (not duplication).

I traced this in prepopulatePlugin.php. The hooks explicitly skip execution when `for_duplication` is true:

if ($this->opo_plugin_config->get('enabled') && !caGetOption('for_duplication', $pa_params, false)) {
    $this->prepopulateFields($pa_params, ['hook' => 'save']);
}

During a normal UI insert/save, I dumped $pa_params and observed that:

caGetOption('for_duplication', $pa_params, false)

was resolving to true, even though no duplication was involved. Because of that, prepopulate correctly skipped execution based on its guard condition.

Tracing it back, the flag originates here:

'for_duplication' => caGetOption('forDuplication', $pa_options, true)

Since the default is true, any insert/update where `forDuplication` is not explicitly set gets treated as duplication.

As a result, this condition in the prepopulate plugin:

!caGetOption('for_duplication', $pa_params, false)

was always false, causing rules to be skipped entirely in our setup.

After changing the default to false:
- Normal insert/update operations are no longer treated as duplication
- Prepopulate rules execute correctly on save
- Actual duplication remains unaffected, as it explicitly sets:
  ['forDuplication' => true]

I verified this both via UI save and via caUtils, where rules behave as expected.

If you prefer, I can scope this more narrowly (e.g. only within the plugin), but from what I can see the default value is the root cause.

collectiveaccess:
So for you prepopulate was *never* firing? 

I don't have a problem with the logic you've enumerated about. We'll test this a bit more before merging however.

Thanks as always for your efforts!

behnaz-zarriz:
To be precise: the prepopulate hooks were firing, but the rules were not executed because the guard condition skipped the call to prepopulateFields().

In our normal UI save case, for_duplication was true in the hook params, so this condition evaluated to false:

!caGetOption('for_duplication', $pa_params, false)

As a result, prepopulateFields() was not called on save.

PR #1912, opened 2026-05-13
Ticket[KULTMMS-1334]: git commit -m "Fix incorrect BCE decade range normalization in browse facet

behnaz-zarriz:
This change fixes incorrect filtering for BCE decade values in normalized date browse facets.

In some cases, TimeExpressionParser returned an open-ended upper bound for BCE decade values (for example: start=-30, end=2021...), causing browse queries to match an excessively large result set.

The fix normalizes invalid BCE decade ranges after parsing and limits them to the expected decade interval.

Scope is intentionally narrow:
- only affects decades normalization
- only applies to BCE ranges with invalid upper bounds
- does not affect CE parsing or SQL generation logic


collectiveaccess:
Resolved this issue by fixing `TimeExpressionParser` to properly resolve BCE decade expressions.

PR #1916, opened 2026-05-19
Ticket[KULTMMS-2724]: Guard missing information service mirror helper in browse facets

behnaz-zarriz:
Prevent fatal errors in object browse filters when
caGetInformationServiceMirrorListInformation() is unavailable.

Some deployed environments contain a BrowseEngine version that expects
information service mirror helper functions which are not present in the
current codebase. This caused browse facets based on attribute/list values
(Object types, Material/Technique, Gender, Inventory status, etc.) to fail
during facet label resolution.

Added defensive function_exists() checks and fallback behavior to preserve
compatibility with older/incomplete deployments while maintaining existing
browse functionality.


collectiveaccess:
If a function is missing it needs to be added.

PR #1917, opened 2026-05-19
Ticket[KULTMMS-2726]: Add missing caGetInformationServiceMirrorListInformation() helper to browseHelpers.php

behnaz-zarriz:
Add missing caGetInformationServiceMirrorListInformation() helper to browseHelpers.php

BrowseEngine.php calls caGetInformationServiceMirrorListInformation()
during browse facet label resolution, but the helper function was not
defined in app/helpers/browseHelpers.php.

This caused browse facets to fail with the following fatal error:

"Fatal error: Uncaught Error: Call to undefined function
caGetInformationServiceMirrorListInformation()"

when loading attribute/list-based browse facets.

Adds the missing helper implementation and restores proper browse
facet rendering for affected deployments.

### DIFF
--- a/app/helpers/browseHelpers.php
+++ b/app/helpers/browseHelpers.php
@@ -317,3 +317,31 @@ function caGetBrowseForType($table, $type=null, ?array $options=null) : ?string 
 	return null;
 }
 # ---------------------------------------
+/**
+ *
+ *
+ *
+ */
+function caGetInformationServiceMirrorListInformation($t_element) {
+	if (!$t_element) {
+		return false;
+	}
+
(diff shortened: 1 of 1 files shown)


collectiveaccess:
What branch were you using? The missing helper is not missing it's in `app/helpers/informationServiceHelpers.php` in the `dev/2.0` branch.

behnaz-zarriz:
We use the master branch. Could you please merge this into master? We should use the master branch because we need the changes it contains.

behnaz-zarriz:
Thanks Seth, after your update (merge with dev/2.0), it was resolved.

PR #1921, opened 2026-05-28
early return timestamps when no user logs

peterjanssens:
we got a (fatal) error when the log_ids are empty, where mysql errors on "wcl.log_id IN ()", not correct syntax for the IN clause.

Can the $timestamps array be early returned before doing the user tampering ?

### DIFF
--- a/app/lib/ApplicationChangeLog.php
+++ b/app/lib/ApplicationChangeLog.php
@@ -702,9 +702,11 @@ class ApplicationChangeLog {
 		}
 		
 		// Add user info
-		$log_ids = array_map(function($v) { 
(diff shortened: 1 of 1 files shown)


collectiveaccess:
I guess you truncated your logs at some point? In that case there wouldn't be timestamps either, but your fix works too.

peterjanssens:
that was indeed an experiment at the time - later I learned that there could be strings attached to the (lengthy-but)-not-just-logs.
I didn't get the error 100% of the times though, so I thought this could be a sane thing to add without having to debug what happens when log-wise (new user without logs, no user, ..).

PR #1925, opened 2026-06-02
This change removes several hardcoded date formatting implementations and aligns date rendering with the configured system-wide date settings.

behnaz-zarriz:
Changes include:
- Use datetime.conf as the primary source for date display formatting.
- Replace hardcoded date() calls with localized date formatting helpers.
- Remove explicit dateFormat overrides in views where the system configuration should be respected.

This ensures consistent date presentation throughout the application and allows administrators to centrally control formatting via `datetime.conf`.

### DIFF
--- a/app/lib/Attributes/Values/DateRangeAttributeValue.php
+++ b/app/lib/Attributes/Values/DateRangeAttributeValue.php
@@ -336,7 +336,10 @@ class DateRangeAttributeValue extends AttributeValue implements IAttributeValue 
 				$vs_date_format = $o_date_config->get('dateFormat');
 			}
 		}
-
+		if ((bool)$va_settings['useDatePicker']) {
+			$vs_date_format = 'delimited';
+		}
+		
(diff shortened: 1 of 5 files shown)


app/lib/Attributes/Values/DateRangeAttributeValue.php
```diff
@@ -336,10 +336,7 @@ public function getDisplayValue($pa_options=null) {
 				$vs_date_format = $o_date_config->get('dateFormat');
 			}
 		}
-		if ((bool)$va_settings['useDatePicker']) {
-			$vs_date_format = 'delimited';
```
collectiveaccess, line 340: The date picker can only handle delimited dates. It will fail otherwise.

behnaz-zarriz: Thanks for the review. I’ve restored the date picker handling in the latest commit.

PR #1929, opened 2026-06-05
follow redirects for aat service

peterjanssens:
AAT failed today because of 301 permanent redirect to the output that curl does not follow by default.

adding 

curl_setopt($o_curl, CURLOPT_FOLLOWLOCATION, true);
curl_setopt($o_curl, CURLOPT_MAXREDIRS, 5);

solved this for our 1.7.x installation(s) but this fix should do the same for the new version (our main fcous is the new version but the above is the plain syntax for 1.7).

The old endpoint then works as is, but maybe we should also investigate what their plans are with the current endpoint ??


collectiveaccess:
What terms are causing a redirect? In regular use I have not seen any 30x returns.

peterjanssens:
we do have alternative queries to AAT like

curl http://vocab.getty.edu/sparql.json?query=SELECT+%3FID+%3FTermPrefLabel+%3FParents+%3FParentsFull+%7B%0D%0A%09%3FID+a+skos%3AConcept%3B+luc%3Aterm+%22water%22%3B+skos%3AinScheme+aat%3A+%3B%0D%0A%09gvp%3AprefLabelGVP+%5Bxl%3AliteralForm+%3FTermPrefLabel%5D.%0D%0A%09%7B%3FID+gvp%3AparentStringAbbrev+%3FParents%7D%0D%0A%09%7B%3FID+gvp%3AparentString+%3FParentsFull%7D%0D%0A%09%7B%3FID+gvp%3AdisplayOrder+%3FOrder%7D%0D%0A%7D+LIMIT+100


browser follows through and returns the json, curl when following returns the lookups fine

PR #1939, opened 2026-06-18
Fix for 1923 movement tracking error

### LINKED ISSUE #1923: [bug] Automatic movement record not created when a storage location is moved
I am using a MacOS Homebrew as a dev environment with Apache, MariaDB, and PHP 8.3. The codebase is Providence dev/2.0 branch. 

I am following the documentation here: 
https://docs.collectiveaccess.org/providence/user/workflow/history_tracking_current_value

I have set `record_movement_information_when_moving_storage_location = 1` and successfully tested a `current_location` history-tracking policy in `...conf/local/app.conf`. When I move a storage location like a box or a cabinet from one room to another, it does not create movement record, and the move is not tracked in the logs of the objects. The box moves successfully, the objects in the box move correctly, the objects move inside to the new location as well, but nothing is updated beside the location: 

- Time stamp on the object does not change from last manual update,
- The object log does not show the change in location. 
- The storage container log updates with a geo-reference change, no other details.
- History tracking – chronology does work to move an object, and shows it in the log 
- Move is not generated, is not findable in search

I did some testing and I think its is in `ca_storage_locations::saveBundlesForScreen()`, Where the movement tracking runs, `$old_parent_id` and `$new_parent_id` are equal on every move, so the guard that skips tracking when there is "no actual movement" fires every time and tracking is never reached. 

See this line of code in `app/models/ca_storage_locations.php`

`if ($old_parent_id === $new_parent_id) { return $rc; }  // don't track if there's no actual movement`

I added some logging and got this result on several move attempts

```
SBFS-DECIDE rc=true pc=true old=8   new=8
SBFS-DECIDE rc=true pc=true old=177 new=177
SBFS-DECIDE rc=true pc=true old=22  new=22

```

### Steps to reproduce
1. Configure current-location history tracking with a `ca_movements` policy element
2. Open a storage location with objects inside
3. Using the `hierarchy_location` bundle, go to the Move tab to select a different parent location.
4. Movement form appears below
5. Fill in the Movement details form and save.

poplarmedia:
Fixes #1923 
Root cause analysis by @Monica-Wood in the issue thread. This PR implements fixes based on her findings.
There are two edits:
- `app/models/ca_storage_locations.php` — fixes `$new_parent_id` to read from the request rather than the model, and moves the guard check to after when the form data is available
- `app/lib/BundlableLabelableBaseModelWithAttributes.php` — prevents `type_id` when set via code from being overwritten by the request


Monica-Wood:
You should really link back to the issue that this is trying to fix, in this case issue #1923.
And if you use someone else's work in a fix, you should at least reference them.  

Not sure you are aware, but you didn't rebase against dev/2.0 before you submitted and it is undoing previous commits in the process, you need to fix that before it can even be considered.

poplarmedia:
@Monica-Wood thank you for double checking my work, and catching the missed rebase. It should be fixed now. My apologies for not crediting your original research in the PR that helped me narrow the problem. That has been addressed as well, as has the ticket link.

PR #1946, opened 2026-07-01
Persist changes made by direct editing and fill handle operations

behnaz-zarriz:
Previously, only changes with the source "edit" were persisted. However, changes made by dragging the Handsontable fill handle use the "autofill" source. As a result, copied values appeared correctly in the table, giving the impression that they had been applied, but they were never persisted and were lost after reloading the page.

This change treats "autofill" changes the same as manual edits by passing each affected cell through the existing save queue. As a result, values entered manually and values applied using the fill handle are now saved consistently.

### DIFF
--- a/assets/ca/ca.tableview.js
+++ b/assets/ca/ca.tableview.js
@@ -271,7 +271,7 @@ var caUI = caUI || {};
 					return cellProperties;
 				},
 				afterChange: function (change, source) {
-					if (source === 'edit') { // only save edits
+                    if (source === 'edit' || source === 'autofill') { // only save edits
 						for(var i in change) {
 							var r = change[i][0];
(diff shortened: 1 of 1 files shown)


collectiveaccess:
What is meant by "dragging the handle"?

behnaz-zarriz:
I mean the Handsontable fill handle. Previously, when dragging the fill handle, only the first or last copied value was actually persisted after saving. The other copied values appeared in the UI but were lost after reloading the page. With this change, all values copied via the fill handle are now saved correctly...
grafik


grafik

collectiveaccess:
Ah. I didn't even realize that was a thing. Nor did any of our users I guess.

behnaz-zarriz:
Thanks Seth.

PR #1949, opened 2026-07-03
Ticket[KULTMMS-2732]: The generated query for the `contains` operator was inconsistent with the query parser.

behnaz-zarriz:
The `contains` operator in the Query Builder did not behave as expected. Instead of performing a contains search, it generated a quoted phrase search.

For example, the Query Builder generated:

`ca_object_lots.idno_stub:"2022"`

which did not return the expected results.

This change generates wildcard queries instead:

`ca_object_lots.idno_stub:*2022*`

which matches the parser behavior and returns the same results as the corresponding SQL query:

```
SELECT COUNT(*)
FROM ca_object_lots
WHERE idno_stub LIKE '%2022%'
  AND deleted = 0;
```

Both the SQL query and the generated search now return the same results(585).

After Change :

grafik

grafik

### DIFF
--- a/assets/ca/ca.querytranslator.js
+++ b/assets/ca/ca.querytranslator.js
@@ -129,7 +129,8 @@ var caUI = caUI || {};
 				case 'begins_with':
 					return prefix + escapeValue('^' + ruleSet.value, true);
 				case 'contains':
-					return prefix + escapeValue(ruleSet.value, true);
+                    // Generate wildcard queries for "contains" to match the parser behavior.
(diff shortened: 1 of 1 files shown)


collectiveaccess:
Just pushed alternate implementation of this that provides similar behavior when enabled in `search.conf`. Option name is `use_substring_search_for_contains_searches`. Please try it out and if it satisfies your requirements close this PR.

behnaz-zarriz:
Thanks Seth, it works. Could you please merge this branch into master?

